Issue #7, opened 2022-09-20
Naive Approach to AToken Snapshotting


apbendi:
When snapshotting, ignore interest earned (your weight is underweight). We could provide a method for users to force a snapshot whenever they wanted.

davidlaprade:
This was done in #11

PR #11, opened 2022-10-09
Aave naive approach

davidlaprade:
RE #7 

Rough usecase:

* there is an ERC20, call it GOV
* GOV is the governance token issued by Protocol
* Protocol inherits from [GovernorCountingFractional](https://github.com/ScopeLift/flexible-voting/blob/master/src/GovernorCountingFractional.sol) and thus will accept fractional voting
* I deposit some GOV into Aave and receive aGOV in return
* aGOV implements [FractionalPool](https://github.com/ScopeLift/flexible-voting/blob/master/src/FractionalPool.sol)-like functionality
* as proposals are made to Protocol, I can express my voting preference on them directly to aGOV
* aGOV will then cast fractionally weighted votes to Protocol according to the expressed voting preferences of its holders

### DIFF
--- a/src/ATokenNaive.sol
+++ b/src/ATokenNaive.sol
@@ -145,6 +145,10 @@ contract ATokenNaive is AToken {
     );
 
     ProposalVote storage _proposalVote = proposalVotes[proposalId];
+    require(
+      _proposalVote.forVotes + _proposalVote.againstVotes + _proposalVote.abstainVotes > 0,
+      "no votes expressed"
+    );
 
     uint256 _proposalSnapshotBlockNumber = governor.proposalSnapshot(proposalId);
 
--- a/src/ATokenNaive.sol
+++ b/src/ATokenNaive.sol
@@ -58,6 +58,9 @@ contract ATokenNaive is AToken {
   /// @notice Map proposalId to an address to whether they have voted on this proposal.
   mapping(uint256 => mapping(address => bool)) private _proposalVotersHasVoted;
 
+  /// @notice Map proposalId to whether or not this contract has cast votes on it.
+  mapping(uint256 => bool) public hasCastVotesOnProposal;
+
   /// @notice Map proposalId to vote totals expressed on this proposal.
   mapping(uint256 => ProposalVote) public proposalVotes;
 
@@ -107,6 +110,10 @@ contract ATokenNaive is AToken {
   /// @param proposalId The proposalId in the associated Governor
   /// @param support The depositor's vote preferences in accordance with the `VoteType` enum.
   function expressVote(uint256 proposalId, uint8 support) external {
+    require(
+      !hasCastVotesOnProposal[proposalId],
+      "too late to express, votes already cast"
+    );
     uint256 weight = getPastDeposits(msg.sender, governor.proposalSnapshot(proposalId));
     require(weight > 0, "no weight");
 
@@ -170,6 +177,7 @@ contract ATokenNaive is AToken {
     // weights. It makes no difference what vote type this is.
     uint8 unusedSupportParam = uint8(VoteType.Abstain);
 
+    hasCastVotesOnProposal[proposalId] = true;
     bytes memory fractionalizedVotes =
       abi.encodePacked(_forVotesToCast, _againstVotesToCast, _abstainVotesToCast);
     governor.castVoteWithReasonAndParams(
--- a/src/ATokenNaive.sol
+++ b/src/ATokenNaive.sol
@@ -89,7 +89,7 @@ contract ATokenNaive is AToken {
   /// @notice Method which returns the deadline (as a block number) by which
   /// depositors must express their voting preferences to this Pool contract. It
   /// will always be before the Governor's corresponding proposal deadline. The
-  /// dealine is exclusive, meaning: if this returns (say) block 424242, then the
+  /// deadline is exclusive, meaning: if this returns (say) block 424242, then the
   /// internal voting period will be over on block 424242. The last block for
   /// internal voting will be 424241.
   /// @param proposalId The ID of the proposal in question.
--- a/src/ATokenNaive.sol
+++ b/src/ATokenNaive.sol
@@ -110,10 +110,7 @@ contract ATokenNaive is AToken {
   /// @param proposalId The proposalId in the associated Governor
   /// @param support The depositor's vote preferences in accordance with the `VoteType` enum.
   function expressVote(uint256 proposalId, uint8 support) external {
-    require(
-      !hasCastVotesOnProposal[proposalId],
-      "too late to express, votes already cast"
-    );
+    require(!hasCastVotesOnProposal[proposalId], "too late to express, votes already cast");
     uint256 weight = getPastDeposits(msg.sender, governor.proposalSnapshot(proposalId));
     require(weight > 0, "no weight");
 
--- a/src/ATokenNaive.sol
+++ b/src/ATokenNaive.sol
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: UNLICENSED
-pragma solidity >=0.8.10;
+pragma solidity 0.8.10;
 
 import {AToken} from "aave-v3-core/contracts/protocol/tokenization/AToken.sol";
 import {Errors} from "aave-v3-core/contracts/protocol/libraries/helpers/Errors.sol";
@@ -13,7 +13,7 @@ import {Math} from "openzeppelin-contracts/contracts/utils/math/Math.sol";
 
 interface IFractionalGovernor {
   function token() external returns (address);
-  function proposalSnapshot(uint256 proposalId) external returns (uint256);
+  function proposalSnapshot(uint256 proposalId) external view returns (uint256);
   function proposalDeadline(uint256 proposalId) external view returns (uint256);
   function castVoteWithReasonAndParams(
     uint256 proposalId,
@@ -71,7 +71,7 @@ contract ATokenNaive is AToken {
 
   /// @dev Constructor.
   /// @param _pool The address of the Pool contract
-  /// @param _governor The address of the flex-voting-compatable governance contract.
+  /// @param _governor The address of the flex-voting-compatible governance contract.
   constructor(IPool _pool, address _governor) AToken(_pool) {
     governor = IFractionalGovernor(_governor);
   }
@@ -122,7 +122,7 @@ contract ATokenNaive is AToken {
   }
 
   /// @notice Causes this contract to cast a vote to the Governor for all the
-  /// tokens it currently holds.  Uses the sum of all depositor voting
+  /// tokens it currently holds. Uses the sum of all depositor voting
   /// expressions to decide how to split its voting weight. Can be called by
   /// anyone, but _must_ be called within `CAST_VOTE_WINDOW` blocks before the
   /// proposal deadline.
@@ -274,7 +274,7 @@ contract ATokenNaive is AToken {
   // This was been copied from OZ's ERC20Votes checkpointing system with minor
   // revisions:
   //   * Replace "Vote" with "Deposit", as deposits are what we need to track
-  //   * Make some variable names longer for readibility
+  //   * Make some variable names longer for readability
   //   * Break lines at 80-characters
   struct Checkpoint {
     uint32 fromBlock;
--- a/src/ATokenNaive.sol
+++ b/src/ATokenNaive.sol
@@ -49,16 +49,11 @@ contract ATokenNaive is AToken {
     uint128 abstainVotes;
   }
 
-  /// @notice Must call castVote within this many blocks of the proposal
-  /// deadline, so-as to allow ample time for all depositors to express their
-  /// vote preferences. Corresponds to roughly 4 hours, assuming 12s/block.
-  uint32 public constant CAST_VOTE_WINDOW = 1200;
-
-  /// @notice Map depositor to deposit amount.
-  mapping(address => uint256) public deposits;
-
-  /// @notice Map borrower to total amount borrowed.
-  mapping(address => uint256) public borrowTotal;
+  /// @notice The number of blocks prior to the proposal deadline within which
+  /// `castVote` may be called. Prior to this window, `castVote` will revert so
+  /// as to give users time to call `expressVote` before votes are sent to the
+  /// governor contract.
+  uint32 public immutable CAST_VOTE_WINDOW;
 
   /// @notice Map proposalId to an address to whether they have voted on this proposal.
   mapping(uint256 => mapping(address => bool)) private _proposalVotersHasVoted;
@@ -66,14 +61,19 @@ contract ATokenNaive is AToken {
   /// @notice Map proposalId to vote totals expressed on this proposal.
   mapping(uint256 => ProposalVote) public proposalVotes;
 
-  /// @notice The governor contract associated with this governance token.
+  /// @notice The governor contract associated with this governance token. It
+  /// must be one that supports fractional voting, e.g.
+  /// GovernorCountingFractional.
   IFractionalGovernor public immutable governor;
 
   /// @dev Constructor.
   /// @param _pool The address of the Pool contract
   /// @param _governor The address of the flex-voting-compatible governance contract.
-  constructor(IPool _pool, address _governor) AToken(_pool) {
+  /// @param _castVoteWindow The number of blocks that users have to express
+  /// their votes on a proposal before votes can be cast.
+  constructor(IPool _pool, address _governor, uint32 _castVoteWindow) AToken(_pool) {
     governor = IFractionalGovernor(_governor);
+    CAST_VOTE_WINDOW = _castVoteWindow;
   }
 
   // TODO Is there a better way to do this? It cannot be done in the constructor
@@ -85,7 +85,10 @@ contract ATokenNaive is AToken {
 
   /// @notice Method which returns the deadline (as a block number) by which
   /// depositors must express their voting preferences to this Pool contract. It
-  /// will always be before the Governor's corresponding proposal deadline.
+  /// will always be before the Governor's corresponding proposal deadline. The
+  /// dealine is inclusive, meaning: if this returns (say) block 424242, then the
+  /// internal voting period will be over on block 424242. The last block for
+  /// internal voting will be 424241.
   /// @param proposalId The ID of the proposal in question.
   function internalVotingPeriodEnd(uint256 proposalId)
     public
@@ -105,9 +108,9 @@ contract ATokenNaive is AToken {
   /// @param support The depositor's vote preferences in accordance with the `VoteType` enum.
   function expressVote(uint256 proposalId, uint8 support) external {
     uint256 weight = getPastDeposits(msg.sender, governor.proposalSnapshot(proposalId));
-    if (weight == 0) revert("no weight");
+    require(weight > 0, "no weight");
 
-    if (_proposalVotersHasVoted[proposalId][msg.sender]) revert("already voted");
+    require(!_proposalVotersHasVoted[proposalId][msg.sender], "already voted");
     _proposalVotersHasVoted[proposalId][msg.sender] = true;
 
     if (support == uint8(VoteType.Against)) {
@@ -125,12 +128,16 @@ contract ATokenNaive is AToken {
   /// tokens it currently holds. Uses the sum of all depositor voting
   /// expressions to decide how to split its voting weight. Can be called by
   /// anyone, but _must_ be called within `CAST_VOTE_WINDOW` blocks before the
-  /// proposal deadline.
+  /// proposal deadline. We don't bother to check if the vote has already been
+  /// cast -- GovernorCountingFractional will revert if it has.
   /// @param proposalId The ID of the proposal which the Pool will now vote on.
   function castVote(uint256 proposalId) external {
-    if (internalVotingPeriodEnd(proposalId) > block.number) revert("cannot castVote yet");
-    uint8 unusedSupportParam = uint8(VoteType.Abstain);
-    ProposalVote memory _proposalVote = proposalVotes[proposalId];
+    require(
+      internalVotingPeriodEnd(proposalId) <= block.number,
+      "cannot castVote during internal voting period"
+    );
+
+    ProposalVote storage _proposalVote = proposalVotes[proposalId];
 
     uint256 _proposalSnapshotBlockNumber = governor.proposalSnapshot(proposalId);
 
@@ -159,6 +166,10 @@ contract ATokenNaive is AToken {
       (_votingWeightAtSnapshot * _proposalVote.abstainVotes) / _totalDepositWeightAtSnapshot
     );
 
+    // This param is ignored by the governor when voting with fractional
+    // weights. It makes no difference what vote type this is.
+    uint8 unusedSupportParam = uint8(VoteType.Abstain);
+
     bytes memory fractionalizedVotes =
       abi.encodePacked(_forVotesToCast, _againstVotesToCast, _abstainVotesToCast);
     governor.castVoteWithReasonAndParams(
@@ -166,12 +177,35 @@ contract ATokenNaive is AToken {
     );
   }
 
+  /// @notice Implements the basic logic to mint a scaled balance token.
+  /// @param caller The address performing the mint
+  /// @param onBehalfOf The address of the user that will receive the scaled tokens
+  /// @param amount The amount of tokens getting minted
+  /// @param index The next liquidity index of the reserve
+  /// @return `true` if the the previous balance of the user was 0
+  function _mintScaledWithCheckpoint(
+    address caller,
+    address onBehalfOf,
+    uint256 amount,
+    uint256 index
+  ) internal returns (bool) {
+    bool _returnVar = _mintScaled(caller, onBehalfOf, amount, index);
+
+    // We increment by `amount` instead of any computed/rebased amounts because
+    // `amount` is what actually gets transferred of the underlying asset. We
+    // need our checkpoints to still match up with the underlying asset balance.
+    _writeCheckpoint(_checkpoints[onBehalfOf], _additionFn, amount);
+    _writeCheckpoint(_totalDepositCheckpoints, _additionFn, amount);
+
+    return _returnVar;
+  }
+
   // forgefmt: disable-start
   //===========================================================================
   // BEGIN: Aave overrides
   //===========================================================================
   /// Note: this has been modified from Aave v3's AToken to call our custom
-  /// mintScaled function.
+  /// mintScaledWithCheckpoint function.
   ///
   /// @inheritdoc IAToken
   function mint(
@@ -180,18 +214,18 @@ contract ATokenNaive is AToken {
     uint256 amount,
     uint256 index
   ) external virtual override onlyPool returns (bool) {
-    return __mintScaled(caller, onBehalfOf, amount, index);
+    return _mintScaledWithCheckpoint(caller, onBehalfOf, amount, index);
   }
 
   /// Note: this has been modified from Aave v3's AToken to call our custom
-  /// mintScaled function.
+  /// mintScaledWithCheckpoint function.
   ///
   /// @inheritdoc IAToken
   function mintToTreasury(uint256 amount, uint256 index) external override onlyPool {
     if (amount == 0) {
       return;
     }
-    __mintScaled(address(POOL), _treasury, amount, index);
+    _mintScaledWithCheckpoint(address(POOL), _treasury, amount, index);
   }
 
   /// Note: this has been modified from Aave v3's AToken to update deposit
@@ -210,7 +244,6 @@ contract ATokenNaive is AToken {
     // We decrement by `amount` instead of any computed/rebased amounts because
     // `amount` is what actually gets transferred of the underlying asset. We
     // need our checkpoints to still match up with the underlying asset balance.
-    deposits[from] -= amount;
     _writeCheckpoint(_checkpoints[from], _subtractionFn, amount);
     _writeCheckpoint(_totalDepositCheckpoints, _subtractionFn, amount);
     // End modifications.
@@ -219,50 +252,6 @@ contract ATokenNaive is AToken {
     if (receiverOfUnderlying != address(this)) {
       IERC20(_underlyingAsset).safeTransfer(receiverOfUnderlying, amount);
     }
-  }
-
-  /// Note: this has been modified from Aave v3's ScaledBalanceTokenBase
-  /// contract to include balance checkpointing code.
-  /// https://github.com/aave/aave-v3-core/blob/f3e037b3638e3b7c98f0c09c56c5efde54f7c5d2/contracts/protocol/tokenization/base/ScaledBalanceTokenBase.sol#L61-L91
-  /// Modifications are as indicated below.
-  ///
-  /// @notice Implements the basic logic to mint a scaled balance token.
-  /// @param caller The address performing the mint
-  /// @param onBehalfOf The address of the user that will receive the scaled tokens
-  /// @param amount The amount of tokens getting minted
-  /// @param index The next liquidity index of the reserve
-  /// @return `true` if the the previous balance of the user was 0
-  function __mintScaled(
-    address caller,
-    address onBehalfOf,
-    uint256 amount,
-    uint256 index
-  ) internal returns (bool) {
-    uint256 amountScaled = amount.rayDiv(index);
-    require(amountScaled != 0, Errors.INVALID_MINT_AMOUNT);
-
-    uint256 scaledBalance = super.balanceOf(onBehalfOf);
-    uint256 balanceIncrease = scaledBalance.rayMul(index) -
-      scaledBalance.rayMul(_userState[onBehalfOf].additionalData);
-
-    _userState[onBehalfOf].additionalData = index.toUint128();
-
-    _mint(onBehalfOf, amountScaled.toUint128());
-
-    // Begin modifications.
-    // We increment by `amount` instead of any computed/rebased amounts because
-    // `amount` is what actually gets transferred of the underlying asset. We
-    // need our checkpoints to still match up with the underlying asset balance.
-    deposits[onBehalfOf] += amount;
-    _writeCheckpoint(_checkpoints[onBehalfOf], _additionFn, amount);
-    _writeCheckpoint(_totalDepositCheckpoints, _additionFn, amount);
-    // End modifications.
-
-    uint256 amountToMint = amount + balanceIncrease;
-    emit Transfer(address(0), onBehalfOf, amountToMint);
-    emit Mint(caller, onBehalfOf, amountToMint, balanceIncrease, index);
-
-    return (scaledBalance == 0);
   }
   //===========================================================================
   // END: Aave overrides
--- a/src/ATokenNaive.sol
+++ b/src/ATokenNaive.sol
@@ -27,7 +27,7 @@ interface IVotingToken {
   function transfer(address to, uint256 amount) external returns (bool);
   function transferFrom(address from, address to, uint256 amount) external returns (bool);
   function delegate(address delegatee) external;
-  function getPastVotes(address account, uint256 blockNumber) external returns (uint256);
+  function getPastVotes(address account, uint256 blockNumber) external view returns (uint256);
 }
 
 contract ATokenNaive is AToken {
@@ -86,7 +86,7 @@ contract ATokenNaive is AToken {
   /// @notice Method which returns the deadline (as a block number) by which
   /// depositors must express their voting preferences to this Pool contract. It
   /// will always be before the Governor's corresponding proposal deadline. The
-  /// dealine is inclusive, meaning: if this returns (say) block 424242, then the
+  /// dealine is exclusive, meaning: if this returns (say) block 424242, then the
   /// internal voting period will be over on block 424242. The last block for
   /// internal voting will be 424241.
   /// @param proposalId The ID of the proposal in question.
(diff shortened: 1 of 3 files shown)


src/ATokenNaive.sol
```diff
@@ -0,0 +1,357 @@
...
+  constructor(IPool _pool, address _governor) AToken(_pool) {
+    governor = IFractionalGovernor(_governor);
+  }
+
+  // TODO Is there a better way to do this? It cannot be done in the constructor
+  // because the AToken is just used a proxy -- it won't share an address with
+  // the implementation (i.e. this code).
+  function selfDelegate() public {
+    IVotingToken(governor.token()).delegate(address(this));
+  }
```
davidlaprade, line 84: Open to suggestions on this

src/ATokenNaive.sol
```diff
@@ -0,0 +1,357 @@
...
+  function internalVotingPeriodEnd(uint256 proposalId)
+    public
+    view
+    returns (uint256 _lastVotingBlock)
+  {
+    _lastVotingBlock = governor.proposalDeadline(proposalId) - CAST_VOTE_WINDOW;
+  }
+
+  /// TODO how to handle onBehalfOf?
+  /// TODO should this revert if the vote has been cast?
```
davidlaprade, line 99: Would like feedback on these two items as well

test/AaveAtokenFork.t.sol
```diff
@@ -277,3 +382,1049 @@ contract AaveAtokenForkTest is Test {
...
+    vm.roll(aToken.internalVotingPeriodEnd(_proposalId) + 1);
+
+    // Submit votes on behalf of the pool.
+    aToken.castVote(_proposalId);
+
+    (uint256 _againstVotes, uint256 _forVotes, uint256 _abstainVotes) =
+      governor.proposalVotes(_proposalId);
+
+    // userA's vote should have beaten userB's.
+    assertGt(_forVotes, _againstVotes, "userA did not have more voting power than userB");
```
davidlaprade, line 1425: This is the test that fails. It's what the naive approach can't handle

src/ATokenNaive.sol
```diff
@@ -0,0 +1,357 @@
...
+      proposalVotes[proposalId].forVotes += SafeCast.toUint128(weight);
+    } else if (support == uint8(VoteType.Abstain)) {
+      proposalVotes[proposalId].abstainVotes += SafeCast.toUint128(weight);
+    } else {
+      revert("invalid support value, must be included in VoteType enum");
+    }
+  }
+
+  /// @notice Causes this contract to cast a vote to the Governor for all the
+  /// tokens it currently holds.  Uses the sum of all depositor voting
```
mds1, line 125:
```suggestion
  /// tokens it currently holds. Uses the sum of all depositor voting
```

src/ATokenNaive.sol
```diff
@@ -0,0 +1,357 @@
...
+  /// expressions to decide how to split its voting weight. Can be called by
+  /// anyone, but _must_ be called within `CAST_VOTE_WINDOW` blocks before the
+  /// proposal deadline.
+  /// @param proposalId The ID of the proposal which the Pool will now vote on.
+  function castVote(uint256 proposalId) external {
```
mds1, line 130: I don't see any checks to prevent this from being called twice while within the window. Does the governor revert in that case, or does it just re-record the votes as a no-op?

src/ATokenNaive.sol
```diff
@@ -0,0 +1,357 @@
...
+  }
+
+  // TODO Is there a better way to do this? It cannot be done in the constructor
+  // because the AToken is just used a proxy -- it won't share an address with
+  // the implementation (i.e. this code).
+  function selfDelegate() public {
+    IVotingToken(governor.token()).delegate(address(this));
+  }
+
+  /// @notice Method which returns the deadline (as a block number) by which
```
mds1, line 86: Recommend clarifying whether this deadline is inclusive or exclusive of the returned number

src/ATokenNaive.sol
```diff
@@ -0,0 +1,357 @@
...
+  /// @notice Causes this contract to cast a vote to the Governor for all the
+  /// tokens it currently holds.  Uses the sum of all depositor voting
+  /// expressions to decide how to split its voting weight. Can be called by
+  /// anyone, but _must_ be called within `CAST_VOTE_WINDOW` blocks before the
+  /// proposal deadline.
+  /// @param proposalId The ID of the proposal which the Pool will now vote on.
+  function castVote(uint256 proposalId) external {
+    if (internalVotingPeriodEnd(proposalId) > block.number) revert("cannot castVote yet");
+    uint8 unusedSupportParam = uint8(VoteType.Abstain);
+    ProposalVote memory _proposalVote = proposalVotes[proposalId];
```
mds1, line 133: Seems there's no reason to copy this to memory, can probably save some gas by changing to `storage`

src/ATokenNaive.sol
```diff
@@ -0,0 +1,357 @@
...
+    address receiverOfUnderlying,
+    uint256 amount,
+    uint256 index
+  ) external virtual override onlyPool {
+    // Begin modifications.
+    //
+    // We decrement by `amount` instead of any computed/rebased amounts because
+    // `amount` is what actually gets transferred of the underlying asset. We
+    // need our checkpoints to still match up with the underlying asset balance.
+    deposits[from] -= amount;
```
mds1, line 213: Does the AToken/Aave not track this elsewhere? I'm surprised we need to do it ourselves. Though actually I can't tell what this is used for, so maybe we don't need this state var?

src/ATokenNaive.sol
```diff
@@ -0,0 +1,357 @@
...
+  // revisions:
+  //   * Replace "Vote" with "Deposit", as deposits are what we need to track
+  //   * Make some variable names longer for readibility
+  //   * Break lines at 80-characters
+  struct Checkpoint {
+    uint32 fromBlock;
+    uint224 deposits;
+  }
+  mapping(address => Checkpoint[]) private _checkpoints;
+  Checkpoint[] private _totalDepositCheckpoints;
```
mds1, line 284: A bit unconventional to have state vars and structs buried in the code like this. Maybe pull out the checkpointing code as an abstract contract you can inherit from if you want to keep this code block together?

src/ATokenNaive.sol
```diff
@@ -0,0 +1,357 @@
...
+  // END: Aave overrides
+  //===========================================================================
+
+  //===========================================================================
+  // BEGIN: Checkpointing code.
+  //===========================================================================
+  // This was been copied from OZ's ERC20Votes checkpointing system with minor
+  // revisions:
+  //   * Replace "Vote" with "Deposit", as deposits are what we need to track
+  //   * Make some variable names longer for readibility
```
mds1, line 277:
```suggestion
  //   * Make some variable names longer for readability
```

Also which ones were make longer? From a quick glance at ERC20Votes I couldn't tell and they seem the same

src/ATokenNaive.sol
```diff
@@ -0,0 +1,357 @@
...
+
+  /// @notice Causes this contract to cast a vote to the Governor for all the
+  /// tokens it currently holds.  Uses the sum of all depositor voting
+  /// expressions to decide how to split its voting weight. Can be called by
+  /// anyone, but _must_ be called within `CAST_VOTE_WINDOW` blocks before the
+  /// proposal deadline.
+  /// @param proposalId The ID of the proposal which the Pool will now vote on.
+  function castVote(uint256 proposalId) external {
+    if (internalVotingPeriodEnd(proposalId) > block.number) revert("cannot castVote yet");
+    uint8 unusedSupportParam = uint8(VoteType.Abstain);
```
mds1, line 132: IIUC, `castVoteWithReasonAndParams` takes a `support` arg which typically is used to inform the governor of your vote, but when fractional voting that parameter is ignored which is what you're doing here, so really you could use any vote type? At first I thought this was implying you can't vote abstain, so must just add a comment to clarify, and/or define this closer to the `castVoteWithReasonAndParams` call

src/ATokenNaive.sol
```diff
@@ -0,0 +1,357 @@
...
+    // We need 256 bits because of the multiplication we're about to do.
+    uint256 _votingWeightAtSnapshot = IVotingToken(address(_underlyingAsset)).getPastVotes(
+      address(this), _proposalSnapshotBlockNumber
+    );
+
+    //      forVotesRaw          forVotesScaled
+    // --------------------- = ---------------------
+    //     totalDeposits        deposits (@snapshot)
+    //
+    // forVotesScaled = forVotesRaw * deposits@snapshot / totalDeposits
```
mds1, line 151: Is this correct: we are forced to vote with the full voting weight, so we need to take the counts and scale them up to the total number of votes since some users may not have voted?

src/ATokenNaive.sol
```diff
@@ -0,0 +1,357 @@
...
+    uint128 _forVotesToCast = SafeCast.toUint128(
+      (_votingWeightAtSnapshot * _proposalVote.forVotes) / _totalDepositWeightAtSnapshot
+    );
+    uint128 _againstVotesToCast = SafeCast.toUint128(
+      (_votingWeightAtSnapshot * _proposalVote.againstVotes) / _totalDepositWeightAtSnapshot
+    );
+    uint128 _abstainVotesToCast = SafeCast.toUint128(
+      (_votingWeightAtSnapshot * _proposalVote.abstainVotes) / _totalDepositWeightAtSnapshot
+    );
```
mds1, line 160: So there's an invariant here that `_forVotesToCast + _againstVotesToCast + _abstainVotesToCast == _votingWeightAtSnapshot`? If so, we should compute abstain by subtracting. Currently all of these multiplications may round down so perhaps you could end up with that invariant not holding? Not sure what would happen in this case.

It looks like `_countVoteFractional` enforces `uint256(forVotes) + againstVotes + abstainVotes <= uint128(weight)` so maybe I was wrong that you have to vote with full weight

src/ATokenNaive.sol
```diff
@@ -0,0 +1,357 @@
...
+    uint256 amountScaled = amount.rayDiv(index);
+    require(amountScaled != 0, Errors.INVALID_MINT_AMOUNT);
+
+    uint256 scaledBalance = super.balanceOf(onBehalfOf);
+    uint256 balanceIncrease = scaledBalance.rayMul(index) -
+      scaledBalance.rayMul(_userState[onBehalfOf].additionalData);
+
+    _userState[onBehalfOf].additionalData = index.toUint128();
+
+    _mint(onBehalfOf, amountScaled.toUint128());
```
mds1, line 250: Why don't we call their `_mintScaled` method here instead of duplicating the logic? Also, maybe we can get more PRs in to make additional methods virtual if it simplifies our implementation

src/ATokenNaive.sol
```diff
@@ -0,0 +1,357 @@
...
+  /// https://github.com/aave/aave-v3-core/blob/f3e037b3638e3b7c98f0c09c56c5efde54f7c5d2/contracts/protocol/tokenization/base/ScaledBalanceTokenBase.sol#L61-L91
+  /// Modifications are as indicated below.
+  ///
+  /// @notice Implements the basic logic to mint a scaled balance token.
+  /// @param caller The address performing the mint
+  /// @param onBehalfOf The address of the user that will receive the scaled tokens
+  /// @param amount The amount of tokens getting minted
+  /// @param index The next liquidity index of the reserve
+  /// @return `true` if the the previous balance of the user was 0
+  function __mintScaled(
```
mds1, line 235: Maybe rename to `_checkpointingMintScaled` or something that better describes the difference between this and `_mintScaled`?

src/ATokenNaive.sol
```diff
@@ -0,0 +1,357 @@
...
+    // End modifications.
+
+    uint256 amountToMint = amount + balanceIncrease;
+    emit Transfer(address(0), onBehalfOf, amountToMint);
+    emit Mint(caller, onBehalfOf, amountToMint, balanceIncrease, index);
+
+    return (scaledBalance == 0);
+  }
+  //===========================================================================
+  // END: Aave overrides
```
mds1, line 268: Nice docs/comments overall, super helpful 👌

mds1 (commented):
Have not reviewed the tests yet, but figured I left enought comments to submit those for now. Looks good and clean overall, nice work

src/ATokenNaive.sol
davidlaprade: > can this logic be in that proxy's initializer method that it calls?

It could be, we'd just need to make that initializer virtual -- [currently it's not](https://github.com/aave/aave-v3-core/blob/f3e037b3638e3b7c98f0c09c56c5efde54f7c5d2/contracts/protocol/tokenization/AToken.sol#L53-L62) -- and then override it in our implementation. Might be worth asking -- I bet they'd let us

> Often you pass arbitrary data to delegatecall on the new implementation to initialize it

Initialization is done by the Aave `PoolConfigurator` [here](https://github.com/aave/aave-v3-core/blob/f3e037b3638e3b7c98f0c09c56c5efde54f7c5d2/contracts/protocol/libraries/logic/ConfiguratorLogic.sol#L245), right after the proxy is deployed. So our hands are tied there.

There is a bit of arbitrary data that we can pass into the initiailzer function, specifically the `bytes params` [argument](https://github.com/aave/aave-v3-core/blob/f3e037b3638e3b7c98f0c09c56c5efde54f7c5d2/contracts/protocol/libraries/types/ConfiguratorInputTypes.sol#L20). The issue is just that the initializer, as written, [does nothing](https://github.com/aave/aave-v3-core/blob/f3e037b3638e3b7c98f0c09c56c5efde54f7c5d2/contracts/protocol/tokenization/AToken.sol#L53-L84) with those bytes.

test/AaveAtokenFork.t.sol
davidlaprade: I updated this test with a comment so that it won't fail anymore

src/ATokenNaive.sol
```diff
@@ -0,0 +1,357 @@
...
+  struct ProposalVote {
+    uint128 againstVotes;
+    uint128 forVotes;
+    uint128 abstainVotes;
+  }
+
+  /// @notice Must call castVote within this many blocks of the proposal
+  /// deadline, so-as to allow ample time for all depositors to express their
+  /// vote preferences. Corresponds to roughly 4 hours, assuming 12s/block.
+  uint32 public constant CAST_VOTE_WINDOW = 1200;
```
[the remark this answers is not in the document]
davidlaprade: Yeah, I agree, it's not ideal.

> this probably should be an immutable at construction

That's a good idea, I'll do that

src/ATokenNaive.sol
```diff
@@ -0,0 +1,357 @@
...
+
+  /// @notice Map depositor to deposit amount.
+  mapping(address => uint256) public deposits;
+
+  /// @notice Map borrower to total amount borrowed.
+  mapping(address => uint256) public borrowTotal;
```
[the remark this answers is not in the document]
davidlaprade: Good catch, added by mistake 👍

src/ATokenNaive.sol
```diff
@@ -0,0 +1,357 @@
...
+  /// @param proposalId The ID of the proposal in question.
+  function internalVotingPeriodEnd(uint256 proposalId)
+    public
+    view
+    returns (uint256 _lastVotingBlock)
+  {
+    _lastVotingBlock = governor.proposalDeadline(proposalId) - CAST_VOTE_WINDOW;
+  }
+
+  /// TODO how to handle onBehalfOf?
```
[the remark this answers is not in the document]
davidlaprade: Do we want to allow people to vote on behalf of others, e.g. if they have an allowance?

src/ATokenNaive.sol
davidlaprade: Sorry, what I meant is: should this revert if `aToken.castVote` has already been called -- e.g. if the aToken has already submitted its votes to the governor for the proposal. If it has, there would be no point for anyone to express their preference to the aToken on that proposal

src/ATokenNaive.sol
```diff
@@ -0,0 +1,357 @@
...
+  /// TODO should this revert if the vote has been cast?
+  /// @notice Allow a depositor to express their voting preference for a given
+  /// proposal. Their preference is recorded internally but not moved to the
+  /// Governor until `castVote` is called. We deliberately do NOT revert if the
+  /// internalVotingPeriodEnd has passed.
```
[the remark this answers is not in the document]
davidlaprade: I think we figured we'd just save everyone else the gas by not bothering to check each time if the aToken had voted on the proposal yet

src/ATokenNaive.sol
davidlaprade: Yeah, the [governor will revert](https://github.com/ScopeLift/flexible-voting/blob/806eb9dd1d59af93053f7aeb6606534671bb02b1/src/GovernorCountingFractional.sol#L101), that's why we don't bother checking. I can add a note about this.

src/ATokenNaive.sol
davidlaprade: > when fractional voting that parameter is ignored which is what you're doing here, so really you could use any vote type

Yup, exactly. Happy to add a comment

src/ATokenNaive.sol
davidlaprade: Great catch!

src/ATokenNaive.sol
davidlaprade: > Is this correct: we are forced to vote with the full voting weight,

No, we aren't forced to. We could vote with less. I think we felt it was just a waste of the votes to not use the entire weight if we could.

src/ATokenNaive.sol
davidlaprade: > If so, we should compute abstain by subtracting

Sure, that's a good point.

src/ATokenNaive.sol
mds1: Actually I think we'd just want to add a new initializer, because we don't want to re-initialize the AToken. Then to actually do the upgrade you'd use this [`upgradeToAndCall`](https://github.com/aave/aave-v3-core/blob/f3e037b3638e3b7c98f0c09c56c5efde54f7c5d2/contracts/protocol/libraries/aave-upgradeability/BaseImmutableAdminUpgradeabilityProxy.sol#L69) method to pass in the new AToken implementation and pass the `data` to delegatecall, which for us might just be calling `selfDelegate`, so we don't necessarily need a new initialization method (and maybe we just ensure that can only be called, to be safe, by reverting if we're already self delegating)

There may be a function they have to bulk upgrade all AToken implementations, but in general that's the method you use to ugprade these type of proxies. (I just went to the [aUSDC token](https://optimistic.etherscan.io/address/0x625E7708f30cA75bfd92586e17077590C60eb4cD#writeContract) on optimistic etherscan and clicked write contract to find that method, since I wasn't sure what proxy pattern they use)

src/ATokenNaive.sol
mds1: Got it. So if `aToken.castVote` is called very early, can other people express their preferences and re-vote? I'm not sure if the governor allows it or if it reverts?

But in general I think yes, revert if it's too late to keep the storage consistency, feels cleaner that way. Do we need a check on `aToken.castVote` to prevent calling until some `internalVotingPeriodStart` timestamp?

src/ATokenNaive.sol
```diff
@@ -0,0 +1,357 @@
...
+  function mintToTreasury(uint256 amount, uint256 index) external override onlyPool {
+    if (amount == 0) {
+      return;
+    }
+    __mintScaled(address(POOL), _treasury, amount, index);
+  }
+
+  /// Note: this has been modified from Aave v3's AToken to update deposit
+  /// balance accordingly. We cannot just call `super` here because the function
+  /// is external.
```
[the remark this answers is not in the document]
davidlaprade: Huh TIL. Okay, if you prefer this I'll just leave as-is

src/ATokenNaive.sol
davidlaprade: > Does the AToken/Aave not track this elsewhere?

It does not appear that they do 🤷 

> maybe we don't need this state var

Oh, right! This is another vestage of the FractionalPool code. We can get rid of it

src/ATokenNaive.sol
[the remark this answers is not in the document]
mds1: Oh wait since these are proxies it can't be immutable, it needs to be a regular storage variable set in the initializer method, plus you may want to change it as the upstream governor upgrades its own params, so probably should have a setter method for it too (in a future PR if you want)

src/ATokenNaive.sol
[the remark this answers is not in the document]
mds1: Sorry, to clarify I think we should call `super.burn` or `AToken.burn` instead of copy/pasting the code

src/ATokenNaive.sol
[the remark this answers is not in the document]
mds1: Got it, that's fair, I'd put that rationale in a comment cause otherwise it does sound like you'd want to revert to keep state internally consistent when a vote ends (i.e. the expressed prefs saved in AToken matching the upstream votes placed is a nice property, bummer to give it up to save gas, but not a big deal + makes sense)

src/ATokenNaive.sol
davidlaprade: I think initially I had embedded this checkpointing logic within their logic because I thought I was going to need a computed amount. But I later realized that I just wanted the `amount` passed in. So I think we can just call their `mintScaled`. Good idea 👍

src/ATokenNaive.sol
davidlaprade: I think I made some variables in the binary search a little longer when I was reading it to help me understand it

src/ATokenNaive.sol
davidlaprade: Fair, let me think about this

src/ATokenNaive.sol
davidlaprade: Done in c324e66

src/ATokenNaive.sol
davidlaprade: Done in c324e66

src/ATokenNaive.sol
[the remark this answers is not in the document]
mds1: Maybe we do the same thing that OZ ERC20Votes does, which I think is that you can vote on behalf of someone else with a sig from them

src/ATokenNaive.sol
mds1: Ah, so if it reverts on subsequent votes, then I think we definitely need this check to prevent casting the votes until some internalVotingPeriodStart block/timestamp is reached? https://github.com/ScopeLift/flexible-voting/pull/11#discussion_r1020549163

src/ATokenNaive.sol
mds1: Then what is the scaling / why isn't it just "submit the exact amount of expressed votes"? The comments say:

> We cannot use the proposalVote numbers alone, since some people with deposits at the snapshot might not have expressed votes.

But if they didn't express votes, wouldn't be in the `proposalVotes` mapping?

src/ATokenNaive.sol
mds1: Per above comment around not submitting full voting weight, this makes sense now and I agree, thanks

src/ATokenNaive.sol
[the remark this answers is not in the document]
mds1: Oh yea, you're right. Clearly I messed up when testing 😅

We can leave as-is for this PR but I think we should lean towards upstream PRs to Aave's codebase so we can avoid copying code as much as possible. For example PR in a `public virtual` change here instead of `external`

src/ATokenNaive.sol
mds1: I think you wrote inclusive but meant exclusive?

>   /// will always be before the Governor's corresponding proposal deadline. The
>   /// dealine is inclusive, meaning: if this returns (say) block 424242, then the
>   /// internal voting period will be over on block 424242. The last block for
>   /// internal voting will be 424241.

This says inclusive, but that if it returns number 424242 then the last block is 424241, which would be excusive?

src/ATokenNaive.sol
davidlaprade: Ohhhh, I see what you mean now. That's actually really nice!

Now that I'm thinking about it, I wonder if the test setup here should actually be:
* deploy a traditional aToken for GOV
* upgrade the aToken to the ATokenNaive implementation (self-delegating in the process)
* run test

My only fear about changing the implementation is that we add some storage slots with our aToken implementation. I _think_ those will just get tacked on at the end (since we're inheriting from their AToken), but I should confirm this. If they mess up storage locations that would be _bad_

src/ATokenNaive.sol
davidlaprade: Ah, yeah, good point

src/ATokenNaive.sol
davidlaprade: > So if aToken.castVote is called very early, can other people express their preferences and re-vote? I'm not sure if the governor allows it or if it reverts?

No, castVote can only be called once. The governor will revert otherwise. That's why we have the explicit  period during which `castVote` _can't_ be called -- it gives people dedicated time to vote. Otherwise someone could just express their voting preference then immediately call `castVote`, effectively burning everyone else's voting power.

> I think yes, revert if it's too late to keep the storage consistency, feels cleaner that way

👍 

> Do we need a check on aToken.castVote to prevent calling until some internalVotingPeriodStart block/timestamp?

That's what [this](https://github.com/ScopeLift/flexible-voting/pull/11/files#diff-5a237783c25a06fff902b80c7117e7e15796237c85e0a05dadc7b888afdcd8ecR135-R138) does.

src/ATokenNaive.sol
davidlaprade: Sorry, I'm confused. We have that check right [here](https://github.com/ScopeLift/flexible-voting/pull/11/files#diff-5a237783c25a06fff902b80c7117e7e15796237c85e0a05dadc7b888afdcd8ecR135-R138)

src/ATokenNaive.sol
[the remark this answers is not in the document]
davidlaprade: Might not be that expensive to just add a `mapping(uint256 => bool) public hasVoted` storage var and write to it in `castVote` then check in here. I'll give it a try and see how much gas it adds

src/ATokenNaive.sol
davidlaprade: We have to scale because people might have borrowed GOV, so the actual voting weight of the aToken is probably less than the number of deposits. So expressed votes have to be scaled down

src/ATokenNaive.sol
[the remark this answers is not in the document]
apbendi: Vote on a behalf is more a "nice to have" feature, and the naive implementation is primarily useful as a proof of concept and demonstrating the test harness all work, etc... Let's make an issue to track it as separate feature to eventually add to the "real" implementations once they're done.

src/ATokenNaive.sol
apbendi: Agreed—revert if the vote has already been cast to avoid confusion.

src/ATokenNaive.sol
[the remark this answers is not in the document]
apbendi: Yeah, as written in the comment above, I think it would be preferable to prevent late voting unless the gas price is significant.

src/ATokenNaive.sol
apbendi: In other words, the scaling is: "Vote with the tokens left in the pool proportional to your AToken holdings"

src/ATokenNaive.sol
apbendi: OZ has a [checkpoints](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/Checkpoints.sol) contract that we may be able to leverage here @davidlaprade.

apbendi (commented):
Looks good on my first pass @davidlaprade, nice work. Left some thoughts on the ongoing discussions.

src/ATokenNaive.sol
mds1: Agree with both using that test setup and verifying we don't shift storage slots !(which, I agree we shouldn't if we inherit from AToken, but verifying that + changing the test setup can be a separate issue/PR)

src/ATokenNaive.sol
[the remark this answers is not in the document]
mds1: @davidlaprade should we have an issue / track somewhere "changes to upstream to AToken", and include this, before marking this resolved? Feel free to resolve this comment yourself once tracking (assuming you agree we should track somewhere)

src/ATokenNaive.sol
mds1: Oops you're right 😅

src/ATokenNaive.sol
mds1: Ok, I get this now. Had to think about it bit more but now I see. Thanks!

src/ATokenNaive.sol
davidlaprade: Done in e629913

src/ATokenNaive.sol
davidlaprade: Tracking here https://github.com/ScopeLift/flexible-voting/issues/14

src/ATokenNaive.sol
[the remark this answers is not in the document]
davidlaprade: Tracking here: https://github.com/ScopeLift/flexible-voting/issues/15

src/ATokenNaive.sol
davidlaprade: Done in 161fd2d

src/ATokenNaive.sol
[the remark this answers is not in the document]
davidlaprade: Implemented in 161fd2d

Without preventing late voting (snapshot at 11822b1):

```
╭─────────────────────────────────────────────────┬─────────────────┬────────┬────────┬────────┬─────────╮
│ src/ATokenNaive.sol:ATokenNaive contract        ┆                 ┆        ┆        ┆        ┆         │
╞═════════════════════════════════════════════════╪═════════════════╪════════╪════════╪════════╪═════════╡
│ Function Name                                   ┆ min             ┆ avg    ┆ median ┆ max    ┆ # calls │
├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┤
│ castVote                                        ┆ 1592            ┆ 54887  ┆ 67815  ┆ 85715  ┆ 35      │
├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┤
│ expressVote                                     ┆ 2443            ┆ 36757  ┆ 47895  ┆ 48521  ┆ 52      │
├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┤
```

Preventing late voting (snapshot at 161fd2d):

```
╭─────────────────────────────────────────────────┬─────────────────┬────────┬────────┬────────┬─────────╮
│ src/ATokenNaive.sol:ATokenNaive contract        ┆                 ┆        ┆        ┆        ┆         │
╞═════════════════════════════════════════════════╪═════════════════╪════════╪════════╪════════╪═════════╡
│ Function Name                                   ┆ min             ┆ avg    ┆ median ┆ max    ┆ # calls │
├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┤
│ castVote                                        ┆ 1570            ┆ 72258  ┆ 88473  ┆ 106373 ┆ 40      │
├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┤
│ expressVote                                     ┆ 739             ┆ 37534  ┆ 50097  ┆ 50758  ┆ 59      │
├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┤

```

Adding late-vote prevention increases `castVote` median gas usage by 30% (20658 gas) and `expressVote` median by ~5% (2202 gas).

In the current gas environment it's not a big deal (~$0.65 assuming 26 gwei/gas, but check my math). Still, I think we probably want to make `castVote` as cheap as possible so as not to disincentivize its being called -- especially if gas + ETH prices go up again. I think not disincentivising `castVote` outweighs better UX for people (or apps) that can't be bothered to check whether or not the aToken has cast its weight before they attempt to express. But curious what you guys think.

Snapshots taken with `$ forge test --gas-report`

src/ATokenNaive.sol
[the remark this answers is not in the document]
davidlaprade: Tracking here https://github.com/ScopeLift/flexible-voting/issues/16

src/ATokenNaive.sol
davidlaprade: Great catch! Tracking this here: https://github.com/ScopeLift/flexible-voting/issues/17

src/ATokenNaive.sol
```diff
@@ -0,0 +1,357 @@
...
+    string calldata reason,
+    bytes memory params
+  ) external returns (uint256);
+}
+
+interface IVotingToken {
+  function transfer(address to, uint256 amount) external returns (bool);
+  function transferFrom(address from, address to, uint256 amount) external returns (bool);
+  function delegate(address delegatee) external;
+  function getPastVotes(address account, uint256 blockNumber) external returns (uint256);
```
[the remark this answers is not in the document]
davidlaprade: Done in [11822b1](https://github.com/ScopeLift/flexible-voting/pull/11/commits/11822b1eca6c9fcebc90ecb4633de6d82b031e79)

src/ATokenNaive.sol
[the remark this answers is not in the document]
davidlaprade: Done in c324e664. Tracking setter in https://github.com/ScopeLift/flexible-voting/issues/18

src/ATokenNaive.sol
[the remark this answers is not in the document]
mds1: Thanks for checking this. Adding ~20k gas makes sense, since it's one SSTORE to a fresh slot, but that is a lot. OTOH that method is only called once per proposal so I don't think we'll be too sensitive to price there, and increasing `expressVote` by 5% (the one where IMO we should be more price sensitive) isn't too bad (but also not trivial).

Maybe we leave for now but create an issue track this discussion to come back to it later? It's probably worth getting opinions from e.g. the Aave team here too

src/ATokenNaive.sol
```diff
@@ -0,0 +1,355 @@
...
+  // because the AToken is just used a proxy -- it won't share an address with
+  // the implementation (i.e. this code).
+  function selfDelegate() public {
+    IVotingToken(governor.token()).delegate(address(this));
+  }
+
+  /// @notice Method which returns the deadline (as a block number) by which
+  /// depositors must express their voting preferences to this Pool contract. It
+  /// will always be before the Governor's corresponding proposal deadline. The
+  /// dealine is exclusive, meaning: if this returns (say) block 424242, then the
```
mds1, line 92:
```suggestion
  /// deadline is exclusive, meaning: if this returns (say) block 424242, then the
```

src/ATokenNaive.sol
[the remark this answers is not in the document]
davidlaprade: Tracking: https://github.com/ScopeLift/flexible-voting/issues/19

apbendi (approved):
Looks good to me @davidlaprade—nice job on this. Getting the test harness in place was obviously a pretty tedious lift, but you did a great job! Will be really helpful as we dive into the two other approaches.

PR #12, opened 2022-11-03
Format the codebase

davidlaprade:
This PR aligns the codebase with ScopeLift's internal style guide for foundry projects. Style guide can be seen [here](https://github.com/ScopeLift/foundry-template). Formatting was done with [scopelint](https://github.com/ScopeLift/scopelint).


davidlaprade:
Nvm, just realized there are some comments that need to be manually reformatted now

.github/workflows/ci.yml
```diff
@@ -0,0 +1,95 @@
+# Copied from: https://github.com/ScopeLift/foundry-template/blob/main/.github/workflows/ci.yml
```
mds1, line 1: Nit, but we can delete this comment IMO. We'll probably adjust the file a bit (e.g. slither preferences, etc) to tailor to the project as things go on, so this may diverge anyway. Feel free to resolve this comment yourself.

foundry.toml
```diff
@@ -1,9 +1,35 @@
...
+  line_length = 100
+  multiline_func_header = "attributes_first"
+  number_underscore = "thousands"
+  quote_style = "double"
+  single_line_statement_blocks = "single"
+  tab_width = 2
+  wrap_comments = true
 
-# See more config options https://github.com/foundry-rs/foundry/tree/master/config
+  # Copied from https://github.com/ScopeLift/foundry-template/blob/main/foundry.toml
```
mds1, line 34: Same nit, we can delete this. Normally you'd create the repo from the template which wouldn't have these comments. Feel free to resolve this comment yourself.

src/GovernorCountingFractional.sol
```diff
@@ -3,6 +3,9 @@
 
 pragma solidity ^0.8.0;
 
+// Disabling forgefmt to stay consistent with OZ's style.
```
mds1, line 6: Is this because we plan to get this upstreamed into OZ?

test/AaveAtokenFork.t.sol
```diff
@@ -1,22 +1,24 @@
...
-import { IPool } from 'aave-v3-core/contracts/interfaces/IPool.sol';
-import { ConfiguratorInputTypes } from 'aave-v3-core/contracts/protocol/libraries/types/ConfiguratorInputTypes.sol';
-import { PoolConfigurator } from 'aave-v3-core/contracts/protocol/pool/PoolConfigurator.sol';
-import { DataTypes } from 'aave-v3-core/contracts/protocol/libraries/types/DataTypes.sol';
-import { AaveOracle } from 'aave-v3-core/contracts/misc/AaveOracle.sol';
-
-import { GovToken } from "./GovToken.sol";
-
-import { Pool } from 'aave-v3-core/contracts/protocol/pool/Pool.sol';
+// forgefmt: disable-start
```
mds1, line 4: Why did you disable formatting for imports? (trying to learn more about how we can improve formatter so these are less commonly needed)

src/GovernorCountingFractional.sol
davidlaprade: Exactly

foundry.toml
```diff
@@ -1,9 +1,35 @@
 [profile.default]
-src = 'src'
-out = 'out'
-libs = ['lib']
+  libs = ['lib']
```
mds1, line 2: the `src`, `out`, and `libs` rows can be deleted since these are just the defaults

foundry.toml
davidlaprade: I actually think it would be nice to include a link to our template in the template, so that when we create new projects from it there is a record of where the code came from. This is the kind of thing that I think could easily be forgotten ("where did these settings come from??"), especially in a long-lived project

.github/workflows/ci.yml
```diff
@@ -0,0 +1,94 @@
+me: CI
+
+on:
+  workflow_dispatch:
+  pull_request:
+  push:
+    branches:
+      - main
```
mds1, line 8: I think CI did not run on this branch because this says `main` but the branch is called `master`? So let's rename one of them

foundry.toml
```diff
@@ -1,9 +1,32 @@
 [profile.default]
-src = 'src'
-out = 'out'
-libs = ['lib']
+  optimizer = true
+  optimizer_runs = 10_000_000
+  solc_version = "0.8.16"
```
mds1, line 4: Does this build for you? I get errors because the Aave contracts are pinned to 0.8.10

```
error[5333]: ParserError: Source file requires different compiler version (current compiler is 0.8.16+commit.07a7930e.Darwin.appleclang) - note that nightly builds are considered to be strictly less than the released version
 --> lib/aave-v3-core/contracts/protocol/tokenization/base/ScaledBalanceTokenBase.sol:2:1:
  |
2 | pragma solidity 0.8.10;
```

src/GovernorCountingFractional.sol
```diff
@@ -175,3 +178,4 @@ abstract contract GovernorCountingFractional is Governor {
         }
     }
 }
+// forgefmt: disable-stop
```
mds1, line 181: Seems `scopelint fmt` suppresses these warnings from `forge fmt`, I should fix that
```
warning: [src/GovernorCountingFractional.sol:181:1] Invalid inline config item: disable-stop
warning: [src/FractionalPool.sol:260:3] Invalid inline config item: disable-stop
warning: [test/AaveAtokenFork.t.sol:21:1] Invalid inline config item: disable-stop
```

foundry.toml
davidlaprade: Haha, no. My bad

davidlaprade: I'm thinking we should just leave out the `solc_version` config for this repo, given that it contains source code that we don't control (and thus solc versions that we don't control). It seems better to just let foundry auto-detect solc version on a case-by-case basis.

.github/workflows/ci.yml
davidlaprade: Done in bab4ade

foundry.toml
davidlaprade: Done in bab4ade

src/GovernorCountingFractional.sol
davidlaprade: Done in bab4ade

foundry.toml
mds1: I'm fine with that, though I do prefer an explicit version (just easier to audit / know what solc version you are using). Feel free to resolve this comment yourself.

.github/workflows/ci.yml
mds1: Huh, why is CI still not running on pushes to this branch? Is it disabled in repo settings or something?

foundry.toml
mds1: Oh, we actually cannot pin a version here, because some files (Aave) require 0.8.10, and others (`FractionaPool.sol` and `FractionaPool.t.sol`) require 0.8.13. So the solutions are:
- Don't pin a version: some files get compiled with 0.8.10, and others with 0.8.17
- Pin 0.8.10 and change our contracts to use 0.8.10

We probably should use 0.8.17 whenever we can, so feel free to resolve this comment yourself

mds1: Pushed f015a53 to clarify this, feel free to tweak phrasing

.github/workflows/ci.yml
mds1: Found the issue 😅 6b40870

mds1 (approved):

Issue #16, opened 2022-11-16
Upstream changes to Aave v3 to make it more extensible


davidlaprade:
We already did this here: https://github.com/aave/aave-v3-core/pull/726

Other functions we should update:
* `aToken.burn` to [make it public](https://github.com/ScopeLift/flexible-voting/pull/11/files#r1020441171) (rather than external)
* [`aToken.intialize`](https://github.com/aave/aave-v3-core/blob/f3e037b3638e3b7c98f0c09c56c5efde54f7c5d2/contracts/protocol/tokenization/AToken.sol#L53-L62) to [make it virtual](https://github.com/ScopeLift/flexible-voting/pull/21/files#r1052489548) (so that we can override flex voting aTokens to self-delegate during initialization)
* [`aToken._transfer`](https://github.com/aave/aave-v3-core/blob/f3e037b3638e3b7c98f0c09c56c5efde54f7c5d2/contracts/protocol/tokenization/AToken.sol#L206-L211) to [make it virtual](https://github.com/ScopeLift/flexible-voting/pull/21#discussion_r1051191099) (so that we can checkpoint during both transfer and transferFrom with the same code)

davidlaprade:
I realized this morning that we don't need to make `aToken.burn()` public to do what we want. See https://github.com/ScopeLift/flexible-voting/pull/21/commits/ffc286730a460d64d763e9075adb853a4faf1046 for a better solution that didn't require any Aave changes.

I pushed a PR to make the other two changes: https://github.com/aave/aave-v3-core/pull/774

apbendi:
@davidlaprade should we close this now that our PR has been merged by Aave?

Issue #14, opened 2022-11-16
Upgrade AToken to Flex Voting AToken within Aave Implementation Test Setup


davidlaprade:
Within aToken implementation test suite `setUp` functions we should:
* deploy GOV to aave
* deploy our new aToken implementation
* upgrade the aGOV token to our new implementation via [`upgradeToAndCall`](https://github.com/aave/aave-v3-core/blob/f3e037b3638e3b7c98f0c09c56c5efde54f7c5d2/contracts/protocol/libraries/aave-upgradeability/BaseImmutableAdminUpgradeabilityProxy.sol#L69)
* self-delegate during [`upgradeToAndCall`](https://github.com/aave/aave-v3-core/blob/f3e037b3638e3b7c98f0c09c56c5efde54f7c5d2/contracts/protocol/libraries/aave-upgradeability/BaseImmutableAdminUpgradeabilityProxy.sol#L69)
* confirm this doesn't change any storage slots

More details in this thread: https://github.com/ScopeLift/flexible-voting/pull/11/files#r1019640978

davidlaprade:
Moved over to the new flex-voting-client repo here: https://github.com/ScopeLift/flexible-voting-clients/issues/7

PR #21, opened 2022-12-13
Flex voting Atoken using scaled balance caching

### LINKED ISSUE #8: Caching Approach for AToken Snapshotting
In relevant AToken methods (mint, burn, transferOnLiquidation, others?) read the relevant ReserveData from the Pool contract after the operation completes and store snapshots of it.

When calculating weight, lookup appropriate snapshot data and reimplement the call to getReserveNormalizedIncome (i.e. getNormalizedIncome under the hood, which is internal) to calculate the balance at the snapshot time.

Potentially complicated, requires code duplication, but provides an exact weight.

davidlaprade:
Fixes #8 

More details about this approach and why I think it works can be found here https://github.com/ScopeLift/flexible-voting/issues/8#issuecomment-1349793292


src/ATokenCheckpointed.sol
```diff
@@ -0,0 +1,307 @@
...
+  /// Note: this has been modified from Aave v3's AToken to call our custom
+  /// mintScaledWithCheckpoint function.
+  ///
+  /// @inheritdoc IAToken
+  function mintToTreasury(uint256 amount, uint256 index) external override onlyPool {
+    if (amount == 0) {
+      return;
+    }
+    _mintScaledWithCheckpoint(address(POOL), _treasury, amount, index);
+  }
```
davidlaprade, line 263: As I'm reviewing this, I actually think we should *not* be checkpointing this function. It's called by the pool when someone calls [`mintToTreasury`](https://github.com/aave/aave-v3-core/blob/f3e037b3638e3b7c98f0c09c56c5efde54f7c5d2/contracts/protocol/pool/Pool.sol#L446-L448).

But mintToTreasury does not involve depositing any underlying tokens into Aave (even though it results in minting aTokens). So there should be no need to checkpoint anything. Adding checkpoints here would actually mess up our internal accounting.

Furthermore, the pool (which is the address that we would be checkpointing for) is not an entity that can vote

src/ATokenCheckpointed.sol
```diff
@@ -0,0 +1,307 @@
...
+    // We need 256 bits because of the multiplication we're about to do.
+    uint256 _votingWeightAtSnapshot = IVotingToken(address(_underlyingAsset)).getPastVotes(
+      address(this), _proposalSnapshotBlockNumber
+    );
+
+    //      forVotesRaw          forVotesScaled
+    // --------------------- = ---------------------
+    //     totalDeposits        deposits (@snapshot)
+    //
+    // forVotesScaled = forVotesRaw * deposits@snapshot / totalDeposits
```
davidlaprade, line 171: This would probably be clearer:

```suggestion
    //      forVotesRaw          forVoteWeight
    // --------------------- = ---------------------
    //     totalDeposits         totalVoteWeight
    //
    // forVoteWeight = forVotesRaw * totalVoteWeight / totalDeposits
```

src/ATokenCheckpointed.sol
```diff
@@ -0,0 +1,307 @@
...
+
+  function _checkpointRawBalanceOf(address _user)
+    internal
+    returns (uint256 _previousBalance, uint256 _currentBalance)
+  {
+    (_previousBalance, _currentBalance) = balanceCheckpoints[_user].push(_rawBalanceOf(_user));
+  }
+
+  // Returns the raw (i.e. unrebased) balanceOf the _user at the _blockNumber.
+  function getPastStoredBalance(address _user, uint256 _blockNumber) public returns (uint256) {
```
davidlaprade, line 228: I need to add full natspec comments to these functions

test/ATokenCheckpointed.t.sol
```diff
@@ -0,0 +1,1974 @@
+// SPDX-License-Identifier: Unlicensed
```
davidlaprade, line 1: Most of these tests were just copied as-is from the naive implementation's tests. Here is a temporary diff to make review easier: https://www.diffchecker.com/A87zLOy3

It's just [this](https://raw.githubusercontent.com/ScopeLift/flexible-voting/7efdd7b8deabd91a4950fc02ab6ef659bac32156/test/ATokenCheckpointed.t.sol) diff'ed against [this](https://raw.githubusercontent.com/ScopeLift/flexible-voting/1cd6967167928a48deb796c2c8f1a2d820a70d31/test/AaveAtokenFork.t.sol).

src/ATokenCheckpointed.sol
```diff
@@ -0,0 +1,307 @@
+// SPDX-License-Identifier: UNLICENSED
```
davidlaprade, line 1: This is very similar to the existing [ATokenNaive](https://github.com/ScopeLift/flexible-voting/blob/1cd6967167928a48deb796c2c8f1a2d820a70d31/src/ATokenNaive.sol) implementation. Here's a diff of the two: https://www.diffchecker.com/WZ9bQlTv

In case that link expires, it's just [this code](https://raw.githubusercontent.com/ScopeLift/flexible-voting/7efdd7b8deabd91a4950fc02ab6ef659bac32156/src/ATokenCheckpointed.sol) diff'ed against [this code](https://raw.githubusercontent.com/ScopeLift/flexible-voting/1cd6967167928a48deb796c2c8f1a2d820a70d31/src/ATokenNaive.sol)

test/ATokenCheckpointed.t.sol
```diff
@@ -0,0 +1,1974 @@
...
+      // forgefmt: disable-end
+    }
+  }
+}
+
+contract GetPastStoredBalanceTest is AaveAtokenForkTest {
+  function test_GetPastStoredBalanceCorrectlyReadsCheckpoints() public {
+    _initiateRebasing();
+
+    address _who = address(0xBEEF);
```
mds1, line 1627: similar comment about `makeAddr` instead

test/ATokenCheckpointed.t.sol
```diff
@@ -0,0 +1,1974 @@
...
+    uint256 _blocksJumped = 42;
+    vm.roll(block.number + _blocksJumped);
+    vm.warp(block.timestamp + 42 days);
+
+    // Rebasing should not affect the raw balance.
+    assertEq(
+      aToken.exposedRawBalanceOf(_who),
+      _rawBalances[0],
+      "raw balance shouldn't be affected by rebasing"
+    );
```
mds1, line 1661: This kind of feels like a no-op assertion. We read a storage slot on L1635, make no state changes (aside from skipping forward in blocks/time), and read the same storage slot again on L1658. Not really a problem I guess, just seems a bit odd since of course the slot won't change 🤷

Maybe it's really just validating that `exposedRawBalanceOf` is reading the right slot/value? Reading `getPastStoredBalance` does make more sense though, since that tests our implementation

test/ATokenCheckpointed.t.sol
```diff
@@ -0,0 +1,1974 @@
...
+
+    // getPastStoredBalance should match the initial raw balance.
+    assertEq(
+      aToken.getPastStoredBalance(_who, block.number - _blocksJumped + 1),
+      _rawBalances[0],
+      "getPastStoredBalance does not match the initial raw balance"
+    );
+
+    // getPastStoredBalance should be able to give us the rebased balance at an
+    // intermediate point.
```
mds1, line 1671: "Rebase balance" to me means "true balance a user sees in their wallet", whereas this is comparing raw balances
```suggestion
    // getPastStoredBalance should be able to give us the raw balance at an
    // intermediate point.
```

test/ATokenCheckpointed.t.sol
```diff
@@ -0,0 +1,1974 @@
...
+    _mintGovAndSupplyToAave(_who, _amountB);
+    _rawBalances[1] = aToken.exposedRawBalanceOf(_who);
+
+    // Advance the clock.
+    uint256 _blocksJumpedSecondTime = 100;
+    vm.roll(block.number + _blocksJumpedSecondTime);
+    vm.warp(block.timestamp + 100 days);
+
+    // Rebasing should not affect the raw balance.
+    assertEq(aToken.exposedRawBalanceOf(_who), _rawBalances[1]);
```
mds1, line 1690: Similar no-op test that will always be true

test/ATokenCheckpointed.t.sol
```diff
@@ -0,0 +1,1974 @@
...
+      _rawBalances[1], // The raw balance pre-withdrawal.
+      aToken.getPastStoredBalance(_who, block.number - _blocksJumpedThirdTime)
+    );
+  }
+
+  function test_GetPastStoredBalanceHandlesTransfers() public {
+    _initiateRebasing();
+
+    address _userA = address(0xBABE);
+    address _userB = address(0xBEEF);
```
mds1, line 1743: similar makeAddr comment

test/ATokenCheckpointed.t.sol
```diff
@@ -0,0 +1,1974 @@
...
+    vm.warp(block.timestamp + 100 days);
+
+    // Get the rebased balance for userA.
+    uint256 _initBalanceUserA = aToken.balanceOf(_userA);
+    uint256 _initBalanceUserB = aToken.balanceOf(_userB);
+    assertEq(_initBalanceUserB, 0);
+
+    // Transfer aTokens to userB.
+    vm.prank(_userA);
+    aToken.transfer(_userB, _initBalanceUserA / 3);
```
mds1, line 1762: Per above, should also have an `aToken.transferFrom` test

test/ATokenCheckpointed.t.sol
```diff
@@ -0,0 +1,1974 @@
...
+}
+
+contract GetPastTotalDepositsTest is AaveAtokenForkTest {
+  function test_GetPastTotalDepositsIncreasesOnDeposit() public {
+    _initiateRebasing();
+    uint256 _initTotalDeposits = INITIAL_REBASING_DEPOSIT;
+    assertEq(aToken.getPastTotalDeposits(block.number - 1), INITIAL_REBASING_DEPOSIT);
+
+    address _userA = address(0xBEEF);
+    address _userB = address(0xFEED);
```
mds1, line 1791: same makeAddr comment

src/ATokenCheckpointed.sol
```diff
@@ -0,0 +1,307 @@
...
+  /// @param _pool The address of the Pool contract
+  /// @param _governor The address of the flex-voting-compatible governance contract.
+  /// @param _castVoteWindow The number of blocks that users have to express
+  /// their votes on a proposal before votes can be cast.
+  constructor(IPool _pool, address _governor, uint32 _castVoteWindow) AToken(_pool) {
+    GOVERNOR = IFractionalGovernor(_governor);
+    CAST_VOTE_WINDOW = _castVoteWindow;
+  }
+
+  // TODO Is there a better way to do this? It cannot be done in the constructor
```
wildmolasses, line 89: nit: it might be more idiomatic to put this in an initializer

src/ATokenCheckpointed.sol
```diff
@@ -0,0 +1,307 @@
...
+  /// @param amount The amount of tokens getting minted
+  /// @param index The next liquidity index of the reserve
+  /// @return `true` if the the previous balance of the user was 0
+  function _mintScaledWithCheckpoint(
+    address caller,
+    address onBehalfOf,
+    uint256 amount,
+    uint256 index
+  ) internal returns (bool) {
+    bool _mintScaledReturn = _mintScaled(caller, onBehalfOf, amount, index);
```
wildmolasses, line 206: do we actually use `_mintScaledReturn`? I see we are returning booleans in a bunch of these mint-related fns, but i'm not sure we're actually consuming these bools anywhere. Docs on AToken say that `_mintScaled` "returns `true` if the the previous balance of the user was 0." In our API is that significant at all? If not, could we return void (or a balance uint instead if useful), dropping the complexity of managing this mint scaled return data? a bit confusing out by the external-facing api too, as `mint` returns this bool. Then again, maybe this is actually really useful bool.

src/ATokenCheckpointed.sol
```diff
@@ -0,0 +1,307 @@
...
+  function _checkpointRawBalanceOf(address _user)
+    internal
+    returns (uint256 _previousBalance, uint256 _currentBalance)
+  {
+    (_previousBalance, _currentBalance) = balanceCheckpoints[_user].push(_rawBalanceOf(_user));
+  }
+
+  // Returns the raw (i.e. unrebased) balanceOf the _user at the _blockNumber.
+  function getPastStoredBalance(address _user, uint256 _blockNumber) public returns (uint256) {
+    // TODO could we use getAtProbablyRecentBlock instead?
```
wildmolasses, line 229: docs for reference: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/Checkpoints.sol#L43-L48

I think `getAtProbablyRecentBlock` seems more appropriate 👍

src/ATokenCheckpointed.sol
```diff
@@ -0,0 +1,307 @@
...
+  /// checkpoint raw balances accordingly.
+  ///
+  /// @inheritdoc IERC20
+  function transfer(address recipient, uint256 amount) external virtual override returns (bool) {
+    uint128 castAmount = amount.toUint128();
+    _transfer(_msgSender(), recipient, castAmount);
+    // Begin modifications.
+    _checkpointRawBalanceOf(_msgSender());
+    _checkpointRawBalanceOf(recipient);
+    // End modifications.
```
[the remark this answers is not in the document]
wildmolasses: ^^ the internal `_transfer` instead

agreed

it seems like there are quite a few internal `_transfer`s! We should be sure to override the right one (aka should checkpoint on both `AToken:_transfer` fns, the one with a validate bool and the one without)

src/ATokenCheckpointed.sol
```diff
@@ -0,0 +1,307 @@
...
+  }
+
+  // Returns the raw (i.e. unrebased) balanceOf the _user at the _blockNumber.
+  function getPastStoredBalance(address _user, uint256 _blockNumber) public returns (uint256) {
+    // TODO could we use getAtProbablyRecentBlock instead?
+    return balanceCheckpoints[_user].getAtBlock(_blockNumber);
+  }
+
+  function getPastTotalDeposits(uint256 _blockNumber) public returns (uint256) {
+    return totalDepositCheckpoints.getAtBlock(_blockNumber);
```
wildmolasses, line 234:
```suggestion
    return totalDepositCheckpoints.getAtProbablyRecentBlock(_blockNumber);
```

same as above

wildmolasses (commented):
Thanks @davidlaprade !

src/ATokenCheckpointed.sol
apbendi: Don't fully understand here either, but this code seems to indicate the treasury is earning some income, so perhaps minting to the treasury is just giving the treasury aTokens for the income it has already earned in the pool (fees?). If that's the case, checkpointing *would* be warranted, I believe.

https://github.com/aave/aave-v3-core/blob/f3e037b3638e3b7c98f0c09c56c5efde54f7c5d2/contracts/protocol/libraries/logic/PoolLogic.sol#L107

src/ATokenCheckpointed.sol
```diff
@@ -0,0 +1,307 @@
...
+  function proposalDeadline(uint256 proposalId) external view returns (uint256);
+  function castVoteWithReasonAndParams(
+    uint256 proposalId,
+    uint8 support,
+    string calldata reason,
+    bytes memory params
+  ) external returns (uint256);
+}
+
+interface IVotingToken {
```
apbendi, line 27: This interface is now used in the Fractional Pool (proof of concept) the Naive AToken, and this AToken. Presumably it will be used by plenty of other use case implementations in the future. Seems reasonable to pull it out into its own file in an Interfaces directory.

Out of scope for this PR, but we should also discuss an updated directory structure for this repo as implementations expand.

src/ATokenCheckpointed.sol
```diff
@@ -0,0 +1,307 @@
...
+  }
+
+  /// @notice Method which returns the deadline (as a block number) by which
+  /// depositors must express their voting preferences to this Pool contract. It
+  /// will always be before the Governor's corresponding proposal deadline. The
+  /// deadline is exclusive, meaning: if this returns (say) block 424242, then the
+  /// internal voting period will be over on block 424242. The last block for
+  /// internal voting will be 424241.
+  /// @param proposalId The ID of the proposal in question.
+  function internalVotingPeriodEnd(uint256 proposalId)
```
apbendi, line 103: Also out of scope for this PR, but just want to put the idea in our heads: a bunch of the infrastructure for internal voting can probably be pulled out into an inheritable abstract contract at some point as well.

src/ATokenCheckpointed.sol
```diff
@@ -0,0 +1,307 @@
...
+
+    // This param is ignored by the governor when voting with fractional
+    // weights. It makes no difference what vote type this is.
+    uint8 unusedSupportParam = uint8(VoteType.Abstain);
+
+    hasCastVotesOnProposal[proposalId] = true;
+    bytes memory fractionalizedVotes =
+      abi.encodePacked(_forVotesToCast, _againstVotesToCast, _abstainVotesToCast);
+    GOVERNOR.castVoteWithReasonAndParams(
+      proposalId, unusedSupportParam, "crowd-sourced vote", fractionalizedVotes
```
apbendi, line 190: Nit: should probably include a better "reason" string, like "Total vote from the AToken Pool" or similar.

src/ATokenCheckpointed.sol
apbendi: Looks like both of these getters should be `view` methods

src/ATokenCheckpointed.sol
```diff
@@ -0,0 +1,307 @@
...
+    uint256 _votingWeightAtSnapshot = IVotingToken(address(_underlyingAsset)).getPastVotes(
+      address(this), _proposalSnapshotBlockNumber
+    );
+
+    //      forVotesRaw          forVotesScaled
+    // --------------------- = ---------------------
+    //     totalDeposits        deposits (@snapshot)
+    //
+    // forVotesScaled = forVotesRaw * deposits@snapshot / totalDeposits
+    uint128 _forVotesToCast = SafeCast.toUint128(
```
apbendi, line 172: Trying to think through rounding repercussions for each of these. Am I right in saying that the way we have it set up, each vote option could be truncated down by 1 wei? Meaning worst case we will only ever end up slightly under the available voting weight, never above. Is that right? Would make for a good invariant test!

src/ATokenCheckpointed.sol
```diff
@@ -0,0 +1,307 @@
...
+
+    (uint256 _preMintRawBalance, uint256 _postMintRawBalance) = _checkpointRawBalanceOf(onBehalfOf);
+    totalDepositCheckpoints.push(
+      totalDepositCheckpoints.latest() + (_postMintRawBalance - _preMintRawBalance)
+    );
+
+    return _mintScaledReturn;
+  }
+
+  function _rawBalanceOf(address _user) internal returns (uint256) {
```
apbendi, line 216: Can be a `view` method

apbendi (commented):
Looks great @davidlaprade, left some additional comments. Mostly minor!

src/ATokenCheckpointed.sol
```diff
@@ -0,0 +1,307 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity 0.8.10;
```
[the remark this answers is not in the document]
davidlaprade: It was a combination of:
* the functionality it tested (i.e. deposit checkpointing) was removed, and
* the new functionality that replaced it (raw balance checkpointing) was tested more thoroughly in the dedicated `GetPastStoredBalanceTest`

src/ATokenCheckpointed.sol
davidlaprade: Good callout, agreed

Created https://github.com/ScopeLift/flexible-voting/issues/22 to do it

src/ATokenCheckpointed.sol
davidlaprade: Good point!

The trouble is that most aTokens will (presumably) already have been initialized if Aave ends up adopting this. So the thinking had been that we'd probably need to do this during an upgrade call. We were tracking that work here: https://github.com/ScopeLift/flexible-voting/issues/14

But you're making me realize that I should update this comment to clarify

src/ATokenCheckpointed.sol
davidlaprade: Yeah, great point. Tracking here: https://github.com/ScopeLift/flexible-voting/issues/23

src/ATokenCheckpointed.sol
davidlaprade: > Meaning worst case we will only ever end up slightly under the available voting weight, never above. Is that right?

Yeah, good question. We're rounding down (via division) in each case, so the _most_ these variables should be able to add up to is `_votingWeightAtSnapshot`

> Would make for a good invariant test!

Agreed! Do you want to make an issue to add invariant tests to the repo?

src/ATokenCheckpointed.sol
davidlaprade: Our code doesn't use the return value. Good point. I figured we shouldn't change the aToken interface though. If Aave upgrades their aTokens to our implementation we don't want it breaking any 3rd party contracts that might rely on the return value.

src/ATokenCheckpointed.sol
davidlaprade: Cool, thanks! I'll make the change

src/ATokenCheckpointed.sol
```diff
@@ -0,0 +1,307 @@
...
+  /// @inheritdoc IAToken
+  function mintToTreasury(uint256 amount, uint256 index) external override onlyPool {
+    if (amount == 0) {
+      return;
+    }
+    _mintScaledWithCheckpoint(address(POOL), _treasury, amount, index);
+  }
+
+  /// Note: this has been modified from Aave v3's AToken to checkpoint raw
+  /// balances. We cannot just call `super` here because the function is external.
```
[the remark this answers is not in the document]
davidlaprade: No updates, but tracked here: https://github.com/ScopeLift/flexible-voting/issues/16

src/ATokenCheckpointed.sol
davidlaprade: > minting to the treasury is just giving the treasury aTokens for the income it has already earned in the pool (fees?). If that's the case, checkpointing would be warranted, I believe

I think that's right: it's minting aTokens to represent the treasury income, which are fees collected at various points, e.g. [here](https://github.com/aave/aave-v3-core/blob/f3e037b3638e3b7c98f0c09c56c5efde54f7c5d2/contracts/protocol/libraries/logic/BridgeLogic.sol#L132) and [here](https://github.com/aave/aave-v3-core/blob/f3e037b3638e3b7c98f0c09c56c5efde54f7c5d2/contracts/protocol/libraries/logic/FlashLoanLogic.sol#L238-L240).

Now that I think more about it, it probably does make sense to include this. We're snapshotting aToken balances now, not deposits. And, while the treasury won't have made any deposits, it will have an aToken balance because of this. If at some point Aave decides it wants to participate in governance with the aTokens it has earned as fees, it should be able to. Thanks for correcting me 👍 

I should probably add a test for this functionality though

src/ATokenCheckpointed.sol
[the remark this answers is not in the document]
davidlaprade: Ahh yeah, great catch, guys!

test/ATokenCheckpointed.t.sol
```diff
@@ -0,0 +1,1974 @@
...
+    assertEq(aToken.balanceOf(_userB), 0);
+
+    // Transfer all aTokens from userA to userB.
+    vm.prank(_userA);
+    aToken.transfer(_userB, _transferAmount);
+    assertEq(aToken.balanceOf(_userA), _weight - _transferAmount);
+    assertEq(aToken.balanceOf(_userB), _transferAmount);
+
+    // Advance one block so that our votes will be checkpointed by the govToken;
+    vm.roll(block.number + 1);
```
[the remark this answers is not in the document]
davidlaprade: Oh, good point. Maybe it's not needed in this test

I had added it because you can't lookup a checkpointed value until after the block has been "mined". But proposing doesn't require looking up checkpointed values. It sets the proposal block, which can later be used to lookup the relevant voting weights.

I'll see where else I can remove this 👍

test/ATokenCheckpointed.t.sol
davidlaprade: Yeah, I don't think we need it. I may have just had it in here while I was TDD-ing the feature

src/ATokenCheckpointed.sol
davidlaprade: Actually, I think you're right, we should add this to the initializer so that new aTokens self-delegate. Existing aTokens will still need to be upgraded, though.

Great catch!

davidlaprade: Dang, `initialize` [isn't virtual](https://github.com/aave/aave-v3-core/blob/f3e037b3638e3b7c98f0c09c56c5efde54f7c5d2/contracts/protocol/tokenization/AToken.sol#L53-L62). This is another upstream change we'll need to make to aave.

For now, I'll add this to our existing ticket for upstream aave changes (https://github.com/ScopeLift/flexible-voting/issues/16) and update the comment accordingly.

src/ATokenCheckpointed.sol
davidlaprade: Done in 9d8171b

src/ATokenCheckpointed.sol
davidlaprade: Done in 9d8171b

src/ATokenCheckpointed.sol
davidlaprade: Comment added in https://github.com/ScopeLift/flexible-voting/commit/9d8171bb0a91c502d19cd1fb1180a7440f578c7f

src/ATokenCheckpointed.sol
davidlaprade: Done in 9d8171bb0a91c502d19cd1fb1180a7440f578c7f

src/ATokenCheckpointed.sol
davidlaprade: Done in 9d8171bb0a91c502d19cd1fb1180a7440f578c7f

src/ATokenCheckpointed.sol
davidlaprade: Thanks! Done in abaa5ff

test/ATokenCheckpointed.t.sol
davidlaprade: Done in abaa5ff

test/ATokenCheckpointed.t.sol
davidlaprade: Done in abaa5ff

src/ATokenCheckpointed.sol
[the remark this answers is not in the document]
davidlaprade: Took care of the compiler warnings in aa6ca4c

test/ATokenCheckpointed.t.sol
davidlaprade: Removed in aa6ca4c

test/ATokenCheckpointed.t.sol
davidlaprade: Removed in aa6ca4c

test/ATokenCheckpointed.t.sol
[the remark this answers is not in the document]
davidlaprade: Removed in f800bc6

test/ATokenCheckpointed.t.sol
davidlaprade: Added in a324d17

src/ATokenCheckpointed.sol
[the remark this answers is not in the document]
davidlaprade: > We should be sure to override the right one (aka should checkpoint on both AToken:_transfer fns, the one with a validate bool and the one without)

I don't think we need to checkpoint in both, since the one without the validation argument [just calls the one with it](https://github.com/aave/aave-v3-core/blob/f3e037b3638e3b7c98f0c09c56c5efde54f7c5d2/contracts/protocol/tokenization/AToken.sol#L239). If we just checkpoint in the latter, that should cover us. But I agree we need to make sure we're overriding the correct function.

Unfortunately I think [the function we want to override](https://github.com/aave/aave-v3-core/blob/f3e037b3638e3b7c98f0c09c56c5efde54f7c5d2/contracts/protocol/tokenization/AToken.sol#L206-L211) isn't virtual. This is another one that we're going to need to get changed upstream. I'll just submit a PR at this point.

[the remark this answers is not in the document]
davidlaprade: This fixes the problem locally (where I made `_transfer/4` virtual): fccbe61

src/ATokenCheckpointed.sol
apbendi: Added https://github.com/ScopeLift/flexible-voting/issues/24

src/ATokenCheckpointed.sol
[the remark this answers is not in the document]
mds1: Hmm I'm still seeing a few, and one error now too (error is from the `_transfer` suggestion, so ignore that for the sake of this convo and we'll move that to the other convo)

```
error[4334]: TypeError: Trying to override non-virtual function. Did you forget to add "virtual"?
   --> lib/aave-v3-core/contracts/protocol/tokenization/AToken.sol:206:3:
    |
206 |   function _transfer(
    |   ^ (Relevant source part starts here and spans across multiple lines).
Note: Overriding function is here:
   --> src/ATokenCheckpointed.sol:308:3:
    |
308 |   function _transfer(
    |   ^ (Relevant source part starts here and spans across multiple lines).



warning[2018]: Warning: Function state mutability can be restricted to view
  --> test/FractionalPool.t.sol:87:3:
   |
87 |   function _commonFuzzerAssumptions(address _address, uint256 _voteWeight, uint8 _supportType)
   |   ^ (Relevant source part starts here and spans across multiple lines).



warning[2018]: Warning: Function state mutability can be restricted to view
   --> test/GovernorCountingFractional.t.sol:172:3:
    |
172 |   function _randomSupportType(uint256 salt) public returns (uint8) {
    |   ^ (Relevant source part starts here and spans across multiple lines).



warning[2018]: Warning: Function state mutability can be restricted to view
   --> test/GovernorCountingFractional.t.sol:176:3:
    |
176 |   function _randomVoteSplit(FractionalVoteSplit memory _voteSplit)
    |   ^ (Relevant source part starts here and spans across multiple lines).

```

test/ATokenCheckpointed.t.sol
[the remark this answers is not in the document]
mds1: There are four other `vm.roll(block.number + 1);` usages, should those also be removed? Tried to test myself, but due to the compiler error I could not test

src/ATokenCheckpointed.sol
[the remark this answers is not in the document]
mds1: >  I'll just submit a PR at this point.

Cool, yea IMO we should prioritize this and then simplify our code. Is your current thinking to keep the `_transfer` override and leave this PR open until the upstream PR is merged, or to revert this `_transfer` change (since it doesn't compile) then handle that in a separate PR? Guess it depends how quick they can merge it upstream

src/ATokenCheckpointed.sol
mds1: Yea agreed with @davidlaprade, we should not change the AToken ABI here as that can break integrator's code

src/ATokenCheckpointed.sol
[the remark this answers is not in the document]
davidlaprade: Yeah, the error is expected. I changed the Aave code locally so that I could override `_transfer` and make the test I wrote pass. The compiler error will continue until we upstream the change and bump the aave version here.

Other warnings fixed in bbfaca1

test/ATokenCheckpointed.t.sol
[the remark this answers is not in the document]
davidlaprade: One was unnecessary, removed it in 53732a8

The others are needed b/c they test our code's checkpointing of raw balances. And the way the OZ checkpointing library works, you can't lookup the checkpointed value at a block until that block is past.

src/ATokenCheckpointed.sol
davidlaprade: We have to make `aToken.initialize` both public and virtual. I did that locally, and then updated the initializer here: d7a6bf5

This will fail in CI (because it doesn't have my local Aave changes) but should pass if/when Aave merges the change

src/ATokenCheckpointed.sol
```diff
@@ -0,0 +1,300 @@
...
+
+  /// Note: this has been modified from Aave v3's MintableIncentivizedERC20 to
+  /// checkpoint raw balances accordingly.
+  ///
+  /// @inheritdoc MintableIncentivizedERC20
+  function _mint(address account, uint128 amount) internal override {
+    MintableIncentivizedERC20._mint(account, amount);
+    _checkpointRawBalanceOf(account);
+    totalDepositCheckpoints.push(totalDepositCheckpoints.latest() + amount);
+  }
```
davidlaprade, line 280: I realized we could simplify our code a bit by overriding `_mint` instead of both `mint` and `mintToTreasury`

src/ATokenCheckpointed.sol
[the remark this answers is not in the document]
davidlaprade: Aave PR is here: https://github.com/aave/aave-v3-core/pull/774

src/ATokenCheckpointed.sol
davidlaprade: Tested in ba4a946

src/ATokenCheckpointed.sol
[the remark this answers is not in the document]
davidlaprade: Aave merged our PR earlier today and I've bumped the dependency in [e3ab13b](https://github.com/ScopeLift/flexible-voting/pull/21/commits/e3ab13b2366ec9238d037d49d2e5016502db31f2). So this should be taken care of now

src/ATokenCheckpointed.sol
davidlaprade: Aave merged our PR earlier today and I've bumped the dependency in [e3ab13b](https://github.com/ScopeLift/flexible-voting/pull/21/commits/e3ab13b2366ec9238d037d49d2e5016502db31f2). So this should be taken care of now

src/ATokenCheckpointed.sol
mds1: @davidlaprade for posterity can you leave a comment here summarizing what `mintToTreasury` is for, when/who calls it, and whether we need to care about it? The commit linked above was rebased away and the most recent comments in this thread are "I think" style comments, e.g. from reading it I'm not certain if that's correct or outdated

test/MockATokenFlexVoting.sol
```diff
@@ -0,0 +1,42 @@
...
+import {IPool} from "aave-v3-core/contracts/interfaces/IPool.sol";
+import {ATokenFlexVoting} from "src/ATokenFlexVoting.sol";
+
+contract MockATokenFlexVoting is ATokenFlexVoting {
+  constructor(IPool _pool, address _governor, uint32 _castVoteWindow)
+    ATokenFlexVoting(_pool, _governor, _castVoteWindow)
+  {}
+
+  function handleRepayment(address user, uint256 amount) external virtual onlyPool {
+    // We need this because the Aave code we compile is ahead of the Aave code deployed on
```
mds1, line 13: Great comments, very helpful

test/MockATokenFlexVoting.sol
```diff
@@ -0,0 +1,42 @@
...
+    // But in the latest Aave v3 code, it is a 3-argument function:
+    //
+    //   https://github.com/aave/aave-v3-core/blob/c38c627683c0db0449b7c9ea2fbd68bde3f8dc01/contracts/protocol/tokenization/AToken.sol#L166-L170
+    //
+    // The change was made as a result of this issue:
+    //
+    //   https://github.com/aave/aave-v3-core/issues/742
+    //
+    // We expect that the on-chain AToken implementation will be upgraded to
+    // AToken.handleRepayment/3 at some point in the future. If/when that happens, we should remove
```
mds1, line 30: Nit, I'd just say "upgraded to the 3-argument version" instead of introducing a new syntax (the `/3`) that's only used once and not explained

test/ATokenFlexVotingFork.t.sol
```diff
@@ -1510,4 +1581,478 @@ contract VoteTest is AaveAtokenForkTest {
...
+    uint256 _initRawBalanceUserA = aToken.exposed_RawBalanceOf(_userA);
+
+    // Advance the clock so that we checkpoint and let some rebasing happen.
+    vm.roll(block.number + 100);
+    vm.warp(block.timestamp + 100 days);
+
+    // Get the rebased balance for userA.
+    uint256 _initBalanceUserA = aToken.balanceOf(_userA);
+    uint256 _initBalanceUserB = aToken.balanceOf(_userB);
+    assertEq(_initBalanceUserB, 0);
```
mds1, line 1809: Comments say "userA" but we only assert for userB. Same inconsistency in previous test

mds1: Also can we assert that user A actually has a balance? Otherwise you can't tell we're not just transferring zero everywhere

test/ATokenFlexVotingFork.t.sol
```diff
@@ -1510,4 +1581,478 @@ contract VoteTest is AaveAtokenForkTest {
...
+
+    // Transfer aTokens to userB.
+    vm.prank(_userA);
+    aToken.approve(address(this), type(uint256).max);
+    aToken.transferFrom(_userA, _userB, _initBalanceUserA / 3);
+    assertApproxEqAbs(aToken.balanceOf(_userA), 2 * _initBalanceUserA / 3, 1);
+    assertApproxEqAbs(aToken.balanceOf(_userB), _initBalanceUserA / 3, 1);
```
mds1, line 1816: Can we use `assertEq` here? We precisely know all numbers involved so should be no rounding errors

test/ATokenFlexVotingFork.t.sol
```diff
@@ -1510,4 +1581,478 @@ contract VoteTest is AaveAtokenForkTest {
...
+
+    // Advance the clock so that we checkpoint.
+    vm.roll(block.number + 1);
+    vm.warp(block.timestamp + 1 days);
+
+    // Confirm voting weight has shifted.
+    assertApproxEqAbs(
```
mds1, line 1823: Similarly, shouldn't these also be exact?

test/ATokenFlexVotingFork.t.sol
davidlaprade: We could, but I think this is easier to understand as a reader of the code

src/ATokenFlexVoting.sol
```diff
@@ -83,15 +86,18 @@ contract ATokenNaive is AToken {
...
 
-  // TODO Is there a better way to do this? It cannot be done in the constructor
-  // because the AToken is just used a proxy -- it won't share an address with
-  // the implementation (i.e. this code).
+  // Self-delegation cannot be done in the constructor because the aToken is
+  // just a proxy -- it won't share an address with the implementation (i.e.
+  // this code). Instead we do it at the end of `initialize`. But even that won't
+  // handle already-initialized aTokens. For those, we'll need to self-delegate
+  // during the upgrade process. More details in this issue:
+  // https://github.com/ScopeLift/flexible-voting/issues/16
```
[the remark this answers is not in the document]
davidlaprade: Done in eea4149

test/MockATokenFlexVoting.sol
davidlaprade: Done in eea4149

test/ATokenFlexVotingFork.t.sol
davidlaprade: Done in eea4149

mds1: The comment is fixed but I don't think the assertion was changed to ensure user A has a balance?

test/ATokenFlexVotingFork.t.sol
mds1: This makes it look like we expect rounding errors though, but we don't

wildmolasses (commented):
a couple questions.

side note, i have not yet audited the tests thoroughly. happy to do so, but didn't want to block this review.

test/ATokenFlexVotingFork.t.sol
```diff
@@ -1510,4 +1581,478 @@ contract VoteTest is AaveAtokenForkTest {
...
+      1 // In case of rounding.
+    );
+    assertApproxEqAbs(
+      aToken.getPastStoredBalance(_userB, block.number - 1),
+      _initRawBalanceUserA / 3,
+      1 // In case of rounding.
+    );
+  }
+
+  function test_MintToTreasuryIsCheckpointed() public {
```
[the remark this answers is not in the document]
davidlaprade: Definitely, that'd be great. The only reason why I'm not fuzzing in any of these tests is because they're fork tests and fuzzing would trigger a ton of RPC calls

test/ATokenFlexVotingFork.t.sol
```diff
@@ -1510,4 +1581,478 @@ contract VoteTest is AaveAtokenForkTest {
...
+}
+
+contract GetPastTotalBalancesTest is AaveAtokenForkTest {
+  function test_GetPastTotalBalancesIncreasesOnDeposit() public {
+    _initiateRebasing();
+    assertEq(aToken.getPastTotalBalances(block.number - 1), INITIAL_REBASING_DEPOSIT);
+
+    address _userA = makeAddr("GetPastTotalBalancesIncreasesOnDeposit _userA");
+    address _userB = makeAddr("GetPastTotalBalancesIncreasesOnDeposit _userB");
+    uint256 _amountA = 4242 ether;
```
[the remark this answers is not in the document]
davidlaprade: Yup, see [above](https://github.com/ScopeLift/flexible-voting/pull/21#discussion_r1060857547)

test/ATokenFlexVotingFork.t.sol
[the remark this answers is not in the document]
mds1: You can fuzz with a deterministic seed so RPC results are cached. I don't think our CI is setup to cache the results so if we go that route let's create an issue / handle it in a separate PR

apbendi (approved):
Looks good to me @davidlaprade. Nice work on this 💪

test/ATokenFlexVotingFork.t.sol
davidlaprade: Oh, sorry, I think I made the change before reading your second comment.

Do we need to assert that? We know that user has an aToken balance b/c the user just supplied to aave

test/ATokenFlexVotingFork.t.sol
davidlaprade: Oh oh, yeah, I see what you mean. Gone 92d97c9

test/ATokenFlexVotingFork.t.sol
davidlaprade: Just added them anyway in 92d97c9

src/ATokenFlexVoting.sol
```diff
@@ -187,94 +195,102 @@ contract ATokenNaive is AToken {
...
-  /// @inheritdoc IAToken
-  function mint(
-    address caller,
-    address onBehalfOf,
-    uint256 amount,
-    uint256 index
-  ) external virtual override onlyPool returns (bool) {
-    return _mintScaledWithCheckpoint(caller, onBehalfOf, amount, index);
+  /// @inheritdoc AToken
+  function initialize(
```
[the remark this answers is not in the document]
davidlaprade: cc @mds1 @apbendi

test/ATokenFlexVotingFork.t.sol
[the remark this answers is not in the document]
davidlaprade: > You can fuzz with a deterministic seed

Doesn't that really limit the value of fuzzing though?

Either way, I created a ticket for it here: https://github.com/ScopeLift/flexible-voting/issues/25

src/ATokenCheckpointed.sol
davidlaprade: Sure!

`mintToTreasury` captures/realizes protocol fees by issuing aTokens to the Aave treasury address.

The function is public, so it can be called by anyone. Ernesto from Aave told us in Telegram that it's called by "somebody in the community I think".

We do need to care about it, for the reason @apbendi stated. ATokens are being issued and the underlying asset could potentially be withdrawn (and used for voting). For consistency, therefore, we should checkpoint when it happens and accord the appropriate voting weight to the Treasury -- even if we think it will never express a voting preference.

We do support this function now, and are checkpointing when it is called, as seen by the test added in ba4a9463270b139cff6009ead572490856a6afc8

src/ATokenFlexVoting.sol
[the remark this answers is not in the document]
apbendi: My gut reaction is to agree with @wildmolasses but I don't feel strongly and I don't think it's worth bikeshedding too much about. Unless @wildmolasses (which I'm guessing he doesn't since he used `nit`) I'd say just go with whatever you'd prefer @davidlaprade.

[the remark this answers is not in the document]
davidlaprade: Thanks. Going to punt for now then, ticket tracking this is here: https://github.com/ScopeLift/flexible-voting/issues/26

wildmolasses (approved):
lgtm!

mds1 (approved):

PR #27, opened 2023-01-06
Atoken miscellaneous cleanup

davidlaprade:
This PR does the following:
* moves Aave overrides so that the initalizer function is near the constructor
* pulls shared interfaces into their own directory to DRY things up
* adds a high-level description of the ATokenFlexVoting contract

### DIFF
--- a/src/ATokenFlexVoting.sol
+++ b/src/ATokenFlexVoting.sol
@@ -18,6 +18,28 @@ import {Math} from "openzeppelin-contracts/contracts/utils/math/Math.sol";
 import {Checkpoints} from "openzeppelin-contracts/contracts/utils/Checkpoints.sol";
 // forgefmt: disable-end
 
+/// @notice This is an extension of Aave V3's AToken contract which makes it possible for AToken
+/// holders to still vote on governance proposals. This way, holders of governance tokens do not
+/// have to choose between earning yeild on Aave and voting. They can do both.
+///
+/// AToken holders are able to call `expressVote` to signal their preference on open governance
+/// proposals. When they do so, this extention records that preference with weight proportional to
+/// the users's AToken balance at the proposal snapshot.
+///
+/// When the proposal deadline nears, the AToken's public `castVote` function is called to roll up
+/// all internal voting records into a single delegated vote to the Governor contract -- a vote
+/// which specifies the exact For/Abstain/Against totals expressed by AToken holders.
+///
+/// This extension has the following requirements:
+///   (a) the underlying token be a governance token
+///   (b) the related governor contract supports flexible voting (see GovernorCountingFractional)
+///
+/// Participating in governance via AToken voting is completely optional. Users otherwise still
+/// supply, borrow, and hold tokens with Aave as usual.
+///
+/// The original AToken that this contract extends is viewable here:
+///
+///   https://github.com/aave/aave-v3-core/blob/master/contracts/protocol/tokenization/AToken.sol
 contract ATokenFlexVoting is AToken {
   using WadRayMath for uint256;
   using SafeCast for uint256;
--- a/src/ATokenFlexVoting.sol
+++ b/src/ATokenFlexVoting.sol
@@ -2,6 +2,8 @@
 pragma solidity 0.8.10;
 
 // forgefmt: disable-start
+import {IFractionalGovernor} from "src/interfaces/IFractionalGovernor.sol";
+import {IVotingToken} from "src/interfaces/IVotingToken.sol";
 import {AToken} from "aave-v3-core/contracts/protocol/tokenization/AToken.sol";
 import {MintableIncentivizedERC20} from "aave-v3-core/contracts/protocol/tokenization/base/MintableIncentivizedERC20.sol";
 import {Errors} from "aave-v3-core/contracts/protocol/libraries/helpers/Errors.sol";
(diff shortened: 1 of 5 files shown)


src/ATokenFlexVoting.sol
```diff
@@ -16,25 +18,28 @@ import {Math} from "openzeppelin-contracts/contracts/utils/math/Math.sol";
...
-interface IVotingToken {
-  function transfer(address to, uint256 amount) external returns (bool);
-  function transferFrom(address from, address to, uint256 amount) external returns (bool);
-  function delegate(address delegatee) external;
-  function getPastVotes(address account, uint256 blockNumber) external view returns (uint256);
-}
-
+/// @notice This is an extension of Aave V3's AToken contract which makes it possible for AToken
+/// holders to still vote on governance proposals. This way, holders of governance tokens do not
+/// have to choose between earning yeild on Aave and voting. They can do both.
```
mds1, line 23:
```suggestion
/// have to choose between earning yield on Aave and voting. They can do both.
```

src/ATokenFlexVoting.sol
```diff
@@ -16,25 +18,28 @@ import {Math} from "openzeppelin-contracts/contracts/utils/math/Math.sol";
...
+///
+/// AToken holders are able to call `expressVote` to signal their preference on open governance
+/// proposals. When they do so, this extention records that preference with weight proportional to
```
mds1, line 26:
```suggestion
/// proposals. When they do so, this extension records that preference with weight proportional to
```

src/ATokenFlexVoting.sol
```diff
@@ -16,25 +18,28 @@ import {Math} from "openzeppelin-contracts/contracts/utils/math/Math.sol";
...
+/// This extension has the following requirements:
+///   (a) the underlying token be a governance token
+///   (b) the related governor contract supports flexible voting (see GovernorCountingFractional)
+///
+/// Participating in governance via AToken voting is completely optional. Users otherwise still
+/// supply, borrow, and hold tokens with Aave as usual.
+///
+/// The original AToken that this contract extends is viewable here:
+///
+///   https://github.com/aave/aave-v3-core/blob/master/contracts/protocol/tokenization/AToken.sol
```
mds1, line 42: This extends it from a particular commit (the commit shown by `git submodule status`), so I'd recommend a permalink to that commit instead of linking to master

src/ATokenFlexVoting.sol
```diff
@@ -2,6 +2,8 @@
 pragma solidity 0.8.10;
 
 // forgefmt: disable-start
+import {IFractionalGovernor} from "src/interfaces/IFractionalGovernor.sol";
+import {IVotingToken} from "src/interfaces/IVotingToken.sol";
```
mds1, line 6: Nit on sort order, `src` should be last, item 2 here: https://book.getfoundry.sh/tutorials/best-practices#general-contract-guidance (should add this to scopelint)

src/FractionalPool.sol
```diff
@@ -1,27 +1,11 @@
 // SPDX-License-Identifier: UNLICENSED
-pragma solidity ^0.8.13;
+pragma solidity ^0.8.10;
 
+import {IFractionalGovernor} from "src/interfaces/IFractionalGovernor.sol";
+import {IVotingToken} from "src/interfaces/IVotingToken.sol";
```
mds1, line 5: Same comment on source order

src/interfaces/IFractionalGovernor.sol
```diff
@@ -0,0 +1,15 @@
+// SPDX-License-Identifier: UNLICENSED
```
mds1, line 1: We should have an issue to review license identifiers, not sure if this is correct but also don't want to block on it

src/interfaces/IFractionalGovernor.sol
```diff
@@ -0,0 +1,15 @@
...
+pragma solidity 0.8.10;
+
+// @notice The interface that flexible voting-compatbile governors are expected to support.
```
mds1, line 4: Need triple slash for natspec
```suggestion
/// @notice The interface that flexible voting-compatbile governors are expected to support.
```

src/interfaces/IVotingToken.sol
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity 0.8.10;
+
+// @notice The interface that flexible voting-compatible voting tokens are expected to support.
```
mds1, line 4:
```suggestion
/// @notice The interface that flexible voting-compatible voting tokens are expected to support.
```

src/ATokenFlexVoting.sol
davidlaprade: Ugh, I always mess this up, thank you!

src/interfaces/IFractionalGovernor.sol
davidlaprade: Tracking in https://github.com/ScopeLift/flexible-voting/issues/28

src/ATokenFlexVoting.sol
davidlaprade: Done in 0695773

src/FractionalPool.sol
davidlaprade: Done in 0695773

mds1 (approved):

apbendi (approved):
Looks good to me! Thanks for cleaning that up @davidlaprade.

PR #29, opened 2023-01-10
Make minor atoken tweaks

davidlaprade:
This PR makes the following small changes to the V3 atoken extension:
* removes two unused libraries
* adds explanation for why we are overriding AToken._transfer
* renames `getPastTotalBalances` (plural) to `getPastTotalBalance` (singular), since there is only one total balance

I noticed these things while I was working on the aave v2 atoken extension.

### DIFF
--- a/src/ATokenFlexVoting.sol
+++ b/src/ATokenFlexVoting.sol
@@ -8,9 +8,7 @@ import {Errors} from "aave-v3-core/contracts/protocol/libraries/helpers/Errors.s
 import {GPv2SafeERC20} from "aave-v3-core/contracts/dependencies/gnosis/contracts/GPv2SafeERC20.sol";
 import {IAToken} from "aave-v3-core/contracts/interfaces/IAToken.sol";
 import {IAaveIncentivesController} from "aave-v3-core/contracts/interfaces/IAaveIncentivesController.sol";
-import {IERC20} from "aave-v3-core/contracts/dependencies/openzeppelin/contracts/IERC20.sol";
 import {IPool} from "aave-v3-core/contracts/interfaces/IPool.sol";
-import {WadRayMath} from "aave-v3-core/contracts/protocol/libraries/math/WadRayMath.sol";
 import {SafeCast} from "openzeppelin-contracts/contracts/utils/math/SafeCast.sol";
 import {Math} from "openzeppelin-contracts/contracts/utils/math/Math.sol";
 import {Checkpoints} from "openzeppelin-contracts/contracts/utils/Checkpoints.sol";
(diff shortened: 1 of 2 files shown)


src/ATokenFlexVoting.sol
```diff
@@ -147,8 +142,10 @@ contract ATokenFlexVoting is AToken {
     totalDepositCheckpoints.push(totalDepositCheckpoints.latest() + amount);
   }
 
-  /// Note: this has been modified from Aave v3's AToken contract to
-  /// checkpoint raw balances accordingly.
+  /// Note: this has been modified from Aave v3's AToken contract to checkpoint raw balances
+  /// accordingly.  Ideally we would have overriden IncentivizedERC20._transfer instead of
+  /// AToken._transfer as we did for _mint and _burn above, but that isn't possible here:
+  /// AToken._transfer *already is* an override of IncentivizedERC20._transfer
```
mds1, line 148: `forge doc` is almost ready so let's make some natspec fixes (maybe other places this needs to be done)
```suggestion
  /// @dev This has been modified from Aave v3's AToken contract to checkpoint raw balances
  /// accordingly.  Ideally we would have overriden `IncentivizedERC20._transfer` instead of
  /// `AToken._transfer` as we did for `_mint` and `_burn`, but that isn't possible here:
  /// `AToken._transfer` *already is* an override of `IncentivizedERC20._transfer`
```

Also this is confusing to me: If `AToken._transfer` *already is* an override of `IncentivizedERC20._transfer`, why is it preferable to override `IncentivizedERC20._transfer`? `ATokenFlexVoting is AToken`, so I'd have guessed we want to override `AToken._transfer` since that's the concrete implementation in the standard AToken flow

mds1 (commented):
LGTM!

mds1 (approved):

src/ATokenFlexVoting.sol
davidlaprade: It is confusing. I got confused myself when I was revisiting this code to write the v2 version.

The existing logic looks like this:
* `AToken._transfer/3` overrides `IncentivizedERC20._transfer` to call `AToken._transfer/4` instead
* `AToken._transfer/4` then calls the original `IncentivizedERC20._transfer`

It would have been ideal to override `IncentivizedERC20._transfer` as that's the level we override `_mint` and `_burn` at. Amounts have already been scaled down by liquidity indexes within those functions. So it's really simple to checkpoint raw balances.

Instead, we override `AToken._transfer/4`. This ensures that all transfers go through our checkpointing code. If we had only overridden `AToken._transfer/3` it would have been possible for someone to skip our checkpointing code (e.g. with [transferOnLiquidation](https://github.com/aave/aave-v3-core/blob/8a3913607df88afcc4b8cc63f3a876ebac7d148b/contracts/protocol/tokenization/AToken.sol#L125)).

Issue #13, opened 2022-11-11
feat: allow accounts to submit multiple, partial votes


mds1:
_This is intended as a discussion, not something that will definitely be implemented._

Using arbitrary `GOV` token and  Aave receipt token `aGOV` for the below.

## Problem

Currently, aGOV token holders vote until `X` blocks before the proposal deadline (the `CAST_VOTE_WINDOW`), to leave time for the rolled up vote to be cast. Block numbers are used because that's how the OZ governor [tracks proposals](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/b2970b96e5e2be297421cd7690e3502e49f7deff/contracts/governance/Governor.sol#L38-L43). This has a few issues:

1. It doesn't work as cleanly on L2s or chains like Avalanche where block production is not at a fixed rate. You have fewer guarantees of how much time will remain to cast the rollup vote.
    1. On chains like Optimisim [where](https://community.optimism.io/docs/developers/build/differences/#block-numbers-and-timestamps) `1 transaction == 1 block`, an attacker can see when `CAST_VOTE_WINDOW` is, and once it's reached, spam transactions to accelerate the block number and end proposal voting sooner, to prevent the rolled up flexible vote from being cast. (Note that in the case of Optimism, once Bedrock is released, it [will be](https://community.optimism.io/docs/developers/build/differences/#address-aliasing-2) a fixed 2 second block time, but the problem still exists for other chains)
2. Voter UX takes a hit, so flexible voting isn't "free" to implement:
    1. Users have no guarantee someone will actually cast their rollup vote unless they send it themselves, so must pay attention after voting to then submit the rolled up transaction if they want to guarantee their vote is cast.
    2. Users have a smaller voting window (i.e. less time to research a proposal and vote on it) when voting with `aGOV` compared to `GOV` voters.
3. Tracking proposal vote counts for an ongoing proposal now lags: Expressed `aGOV` votes are not cast until near the deadline, so other voters have less information about existing votes, and there may be a large, last minute change once the rollup vote is cast. You can solve this with a UI/data layer that aggregates across all flexible voting contracts, but (1) that's complex, and (2) there's no guarantee those votes will all be cast.

## Solution

Modify `GovernorCountingFractional` so accounts can vote multiple times, but cannot re-cast votes. Assuming I have 10 GOV tokens, here are four scenarios illustrating UX that is possible (ignoring the `abstain` option for simplicity in the examples):

[...]

Issue #18, opened 2022-11-16
Add setter method for AToken CAST_VOTE_WINDOW


davidlaprade:
RE https://github.com/ScopeLift/flexible-voting/pull/11#discussion_r1020557211

apbendi:
One possible way this might ultimately get handled is for this parameter to be a virtual one on an interface defined as part of #23. Then any concrete use case can set their own value as they see fit.

PR #31, opened 2023-01-19
Make GovernorCountingFractional accept partial votes

davidlaprade:
Implements #13

### DIFF
--- a/src/GovernorCountingFractional.sol
+++ b/src/GovernorCountingFractional.sol
@@ -99,7 +99,7 @@ abstract contract GovernorCountingFractional is Governor {
         bytes memory voteData
     ) internal virtual override {
         require(weight > 0, "GovernorCountingFractional: no weight");
-        if (_proposalVotersWeightCast[proposalId][account] == weight) {
+        if (_proposalVotersWeightCast[proposalId][account] >= weight) {
           revert("GovernorCountingFractional: all weight cast");
         }
 
--- a/src/GovernorCountingFractional.sol
+++ b/src/GovernorCountingFractional.sol
@@ -121,6 +121,11 @@ abstract contract GovernorCountingFractional is Governor {
         uint128 weight,
         uint8 support
     ) internal {
+        require(
+            _proposalVotersWeightCast[proposalId][account] == 0,
+            "GovernorCountingFractional: vote would exceed weight"
+        );
+
         _proposalVotersWeightCast[proposalId][account] = weight;
 
         if (support == uint8(GovernorCompatibilityBravo.VoteType.Against)) {
@@ -147,20 +152,22 @@ abstract contract GovernorCountingFractional is Governor {
     ) internal {
         require(voteData.length == 48, "GovernorCountingFractional: invalid voteData");
 
-        (uint128 forVotes, uint128 againstVotes, uint128 abstainVotes) = _decodePackedVotes(voteData);
+        (uint128 _forVotes, uint128 _againstVotes, uint128 _abstainVotes) = _decodePackedVotes(voteData);
 
-        uint128 remainingWeight = weight - _proposalVotersWeightCast[proposalId][account];
+        uint128 _existingWeight = _proposalVotersWeightCast[proposalId][account];
 
         require(
-            uint256(forVotes) + againstVotes + abstainVotes <= remainingWeight,
-            "GovernorCountingFractional: votes exceed remaining weight"
+            uint256(_forVotes) + _againstVotes + _abstainVotes + _existingWeight <= weight,
+            "GovernorCountingFractional: vote would exceed weight"
         );
+
+        _proposalVotersWeightCast[proposalId][account] = _existingWeight + _forVotes + _againstVotes + _abstainVotes;
 
         ProposalVote memory _proposalVote = _proposalVotes[proposalId];
         _proposalVote = ProposalVote(
-            _proposalVote.againstVotes + againstVotes,
-            _proposalVote.forVotes + forVotes,
-            _proposalVote.abstainVotes + abstainVotes
+            _proposalVote.againstVotes + _againstVotes,
+            _proposalVote.forVotes + _forVotes,
+            _proposalVote.abstainVotes + _abstainVotes
         );
 
         _proposalVotes[proposalId] = _proposalVote;
--- a/src/GovernorCountingFractional.sol
+++ b/src/GovernorCountingFractional.sol
@@ -98,11 +98,6 @@ abstract contract GovernorCountingFractional is Governor {
         uint256 weight,
         bytes memory voteData
     ) internal virtual override {
-        require(
-          _proposalVotersWeightCast[proposalId][account] < weight,
-          "GovernorCountingFractional: vote would exceed weight"
-        );
-
         uint128 safeWeight = SafeCast.toUint128(weight);
 
         if (voteData.length == 0) {
--- a/src/GovernorCountingFractional.sol
+++ b/src/GovernorCountingFractional.sol
@@ -26,7 +26,16 @@ abstract contract GovernorCountingFractional is Governor {
         uint128 abstainVotes;
     }
 
+    /**
+     * @dev Mapping from proposal ID to vote tallies for that proposal.
+     */
     mapping(uint256 => ProposalVote) private _proposalVotes;
+
+    /**
+     * @dev Mapping from proposal ID and address to the weight the address
+     * has cast on that proposal, e.g. _proposalVotersWeightCast[42][0xBEEF]
+     * would tell you the number of votes that 0xBEEF has cast on proposal 42.
+     */
     mapping(uint256 => mapping(address => uint128)) private _proposalVotersWeightCast;
 
     /**
--- a/src/GovernorCountingFractional.sol
+++ b/src/GovernorCountingFractional.sol
@@ -98,6 +98,11 @@ abstract contract GovernorCountingFractional is Governor {
         uint256 weight,
         bytes memory voteData
     ) internal virtual override {
+        require(weight > 0, "GovernorCountingFractional: no weight");
+        if (_proposalVotersWeightCast[proposalId][account] == weight) {
+          revert("GovernorCountingFractional: vote already cast");
+        }
+
         uint128 safeWeight = SafeCast.toUint128(weight);
 
         if (voteData.length == 0) {
--- a/src/GovernorCountingFractional.sol
+++ b/src/GovernorCountingFractional.sol
@@ -27,7 +27,7 @@ abstract contract GovernorCountingFractional is Governor {
     }
 
     mapping(uint256 => ProposalVote) private _proposalVotes;
-    mapping(uint256 => mapping(address => bool)) private _proposalVotersHasVoted;
+    mapping(uint256 => mapping(address => uint128)) private _proposalVotersWeightCast;
 
     /**
      * @dev See {IGovernor-COUNTING_MODE}.
@@ -41,7 +41,7 @@ abstract contract GovernorCountingFractional is Governor {
      * @dev See {IGovernor-hasVoted}.
      */
     function hasVoted(uint256 proposalId, address account) public view virtual override returns (bool) {
-        return _proposalVotersHasVoted[proposalId][account];
+        return _proposalVotersWeightCast[proposalId][account] > 0;
     }
 
     /**
@@ -98,13 +98,17 @@ abstract contract GovernorCountingFractional is Governor {
         uint256 weight,
         bytes memory voteData
     ) internal virtual override {
-        require(!_proposalVotersHasVoted[proposalId][account], "GovernorCountingFractional: vote already cast");
-        _proposalVotersHasVoted[proposalId][account] = true;
+        require(
+          _proposalVotersWeightCast[proposalId][account] < weight,
+          "GovernorCountingFractional: vote would exceed weight"
+        );
+
+        uint128 safeWeight = SafeCast.toUint128(weight);
 
         if (voteData.length == 0) {
-            _countVoteNominal(proposalId, support, weight);
+            _countVoteNominal(proposalId, account, safeWeight, support);
         } else {
-            _countVoteFractional(proposalId, weight, voteData);
+            _countVoteFractional(proposalId, account, safeWeight, voteData);
         }
     }
 
@@ -113,15 +117,18 @@ abstract contract GovernorCountingFractional is Governor {
      */
     function _countVoteNominal(
         uint256 proposalId,
-        uint8 support,
-        uint256 weight
+        address account,
+        uint128 weight,
+        uint8 support
     ) internal {
+        _proposalVotersWeightCast[proposalId][account] = weight;
+
         if (support == uint8(GovernorCompatibilityBravo.VoteType.Against)) {
-            _proposalVotes[proposalId].againstVotes += SafeCast.toUint128(weight);
+            _proposalVotes[proposalId].againstVotes += weight;
         } else if (support == uint8(GovernorCompatibilityBravo.VoteType.For)) {
-            _proposalVotes[proposalId].forVotes += SafeCast.toUint128(weight);
+            _proposalVotes[proposalId].forVotes += weight;
         } else if (support == uint8(GovernorCompatibilityBravo.VoteType.Abstain)) {
-            _proposalVotes[proposalId].abstainVotes += SafeCast.toUint128(weight);
+            _proposalVotes[proposalId].abstainVotes += weight;
         } else {
             revert("GovernorCountingFractional: invalid support value, must be included in VoteType enum");
         }
@@ -134,16 +141,19 @@ abstract contract GovernorCountingFractional is Governor {
      */
     function _countVoteFractional(
         uint256 proposalId,
-        uint256 weight,
+        address account,
+        uint128 weight,
         bytes memory voteData
     ) internal {
         require(voteData.length == 48, "GovernorCountingFractional: invalid voteData");
 
         (uint128 forVotes, uint128 againstVotes, uint128 abstainVotes) = _decodePackedVotes(voteData);
 
+        uint128 remainingWeight = weight - _proposalVotersWeightCast[proposalId][account];
+
         require(
-            uint256(forVotes) + againstVotes + abstainVotes <= uint128(weight),
-            "GovernorCountingFractional: votes exceed weight"
+            uint256(forVotes) + againstVotes + abstainVotes <= remainingWeight,
+            "GovernorCountingFractional: votes exceed remaining weight"
         );
 
         ProposalVote memory _proposalVote = _proposalVotes[proposalId];
--- a/src/GovernorCountingFractional.sol
+++ b/src/GovernorCountingFractional.sol
@@ -57,8 +57,8 @@ abstract contract GovernorCountingFractional is Governor {
             uint256 abstainVotes
         )
     {
-        ProposalVote storage proposalvote = _proposalVotes[proposalId];
-        return (proposalvote.againstVotes, proposalvote.forVotes, proposalvote.abstainVotes);
+        ProposalVote storage proposalVote = _proposalVotes[proposalId];
+        return (proposalVote.againstVotes, proposalVote.forVotes, proposalVote.abstainVotes);
     }
 
     /**
@@ -100,7 +100,7 @@ abstract contract GovernorCountingFractional is Governor {
     ) internal virtual override {
         require(weight > 0, "GovernorCountingFractional: no weight");
         if (_proposalVotersWeightCast[proposalId][account] == weight) {
-          revert("GovernorCountingFractional: vote already cast");
+          revert("GovernorCountingFractional: all weight cast");
         }
 
         uint128 safeWeight = SafeCast.toUint128(weight);
@@ -155,13 +155,13 @@ abstract contract GovernorCountingFractional is Governor {
         (uint128 _forVotes, uint128 _againstVotes, uint128 _abstainVotes) = _decodePackedVotes(voteData);
 
         uint128 _existingWeight = _proposalVotersWeightCast[proposalId][account];
+        uint256 _newWeight = uint256(_forVotes) + _againstVotes + _abstainVotes + _existingWeight;
 
-        require(
-            uint256(_forVotes) + _againstVotes + _abstainVotes + _existingWeight <= weight,
-            "GovernorCountingFractional: vote would exceed weight"
-        );
+        require(_newWeight <= weight, "GovernorCountingFractional: vote would exceed weight");
 
-        _proposalVotersWeightCast[proposalId][account] = _existingWeight + _forVotes + _againstVotes + _abstainVotes;
+        // It's safe to downcast here because we've just confirmed that
+        // _newWeight < weight, and we know that weight is <= uint128.max.
+        _proposalVotersWeightCast[proposalId][account] = uint128(_newWeight);
 
         ProposalVote memory _proposalVote = _proposalVotes[proposalId];
         _proposalVote = ProposalVote(
--- a/src/GovernorCountingFractional.sol
+++ b/src/GovernorCountingFractional.sol
@@ -6,9 +6,9 @@ pragma solidity ^0.8.0;
 // Disabling forgefmt to stay consistent with OZ's style.
 // forgefmt: disable-start
 
-import "openzeppelin-contracts/governance/Governor.sol";
-import "openzeppelin-contracts/governance/compatibility/GovernorCompatibilityBravo.sol";
-import "openzeppelin-contracts/utils/math/SafeCast.sol";
+import {Governor} from "openzeppelin-contracts/governance/Governor.sol";
+import {GovernorCompatibilityBravo} from "openzeppelin-contracts/governance/compatibility/GovernorCompatibilityBravo.sol";
+import {SafeCast} from "openzeppelin-contracts/utils/math/SafeCast.sol";
 
 /**
  * @notice Extension of {Governor} for 3 option fractional vote counting. When voting, a delegate may split their vote
--- a/test/GovernorCountingFractional.t.sol
+++ b/test/GovernorCountingFractional.t.sol
@@ -44,6 +44,8 @@ contract GovernorCountingFractionalTest is Test {
   // This is the vote storage slot size on the Fractional Governor contract.
   uint256 constant MAX_VOTE_WEIGHT = type(uint128).max;
 
+  bytes32 internal nextUser = keccak256(abi.encodePacked("user address"));
+
   struct FractionalVoteSplit {
     uint256 percentFor; // wad
     uint256 percentAgainst; // wad
@@ -142,11 +144,11 @@ contract GovernorCountingFractionalTest is Test {
   function _executeProposal() internal {
     Proposal memory _rawProposalInfo = _getSimpleProposal();
 
-    vm.expectEmit(true, false, false, false);
+    vm.expectEmit(true, true, true, true);
     emit ProposalExecuted(_rawProposalInfo.id);
 
     // Ensure that the other contract is invoked.
-    vm.expectEmit(false, false, false, false);
+    vm.expectEmit(true, true, true, true);
     emit MockFunctionCalled();
 
     governor.execute(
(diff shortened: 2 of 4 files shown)


src/GovernorCountingFractional.sol
```diff
@@ -134,23 +146,28 @@ abstract contract GovernorCountingFractional is Governor {
...
         uint256 proposalId,
-        uint256 weight,
+        address account,
+        uint128 weight,
         bytes memory voteData
     ) internal {
         require(voteData.length == 48, "GovernorCountingFractional: invalid voteData");
 
-        (uint128 forVotes, uint128 againstVotes, uint128 abstainVotes) = _decodePackedVotes(voteData);
+        (uint128 _forVotes, uint128 _againstVotes, uint128 _abstainVotes) = _decodePackedVotes(voteData);
```
mds1, line 155: _(arbitrary comment location to start a thread)_

One small comment is that this PR would have been easier to review without the variable name changes (i.e. adding a leading underscore), since lots of lines had changes like this so I had to see if that was the only change on the line or not. Ideally we should handle things like that in their own PR to keep diffs simpler, not a big deal though of course 🙂

src/GovernorCountingFractional.sol
```diff
@@ -134,23 +146,28 @@ abstract contract GovernorCountingFractional is Governor {
...
+        uint128 _existingWeight = _proposalVotersWeightCast[proposalId][account];
 
         require(
-            uint256(forVotes) + againstVotes + abstainVotes <= uint128(weight),
-            "GovernorCountingFractional: votes exceed weight"
+            uint256(_forVotes) + _againstVotes + _abstainVotes + _existingWeight <= weight,
+            "GovernorCountingFractional: vote would exceed weight"
         );
 
+        _proposalVotersWeightCast[proposalId][account] = _existingWeight + _forVotes + _againstVotes + _abstainVotes;
```
mds1, line 164: Could save that sum to a var instead of computing it twice

mds1 (commented):
looks great!

test/GovernorCountingFractional.t.sol
```diff
@@ -165,8 +171,18 @@ contract GovernorCountingFractionalTest is Test {
...
+  {
+    vm.assume(_addr > address(0));
+    assumeNoPrecompiles(_addr);
+    vm.label(_addr, _name);
+    return _addr;
+  }
+
   function _randomAddress(uint256 salt1) public view returns (address) {
-    return address(uint160(uint256(keccak256(abi.encodePacked(salt1, block.number)))));
+    return address(uint160(uint256(keccak256(abi.encodePacked(salt1, block.timestamp)))));
```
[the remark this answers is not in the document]
davidlaprade: Oh, interesting, I didn't realize that `block.timestamp` was locked in tests. Thanks, I like that approach

test/GovernorCountingFractional.t.sol
```diff
@@ -249,35 +265,37 @@ contract GovernorCountingFractionalTest is Test {
...
 
-  function _castVotes(Voter[4] memory voters, uint256 _proposalId) internal {
-    Voter memory voter;
-    for (uint8 _i = 0; _i < voters.length; _i++) {
-      voter = voters[_i];
-
-      assert(!governor.hasVoted(_proposalId, voter.addr));
+  function _castVotes(Voter memory _voter, uint256 _proposalId) internal {
+    if (_voter.weight == 0) return;
+    assert(!governor.hasVoted(_proposalId, _voter.addr));
```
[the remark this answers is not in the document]
davidlaprade: To be honest, I don't remember exactly what I wanted when I added these -- it was a while ago. I'll revisit

test/GovernorCountingFractional.t.sol
```diff
@@ -454,33 +499,42 @@ contract GovernorCountingFractionalTest is Test {
     voter.addr = _randomAddress(_weight);
     // The weight cannot overflow the max supply for the token, but must overflow the
     // max for the GovernorFractional contract.
-    voter.weight = bound(_weight, MAX_VOTE_WEIGHT, token.exposed_maxSupply());
+    voter.weight = bound(_weight, MAX_VOTE_WEIGHT + 1, token.exposed_maxSupply());
```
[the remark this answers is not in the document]
davidlaprade: Per the comment above, we need a weight that is greater than the max voting weight for the governor

test/GovernorCountingFractional.t.sol
```diff
@@ -454,33 +499,42 @@ contract GovernorCountingFractionalTest is Test {
...
   function testFuzz_ParamLengthIsChecked(
     uint256 _weight,
     FractionalVoteSplit memory _voteSplit,
-    uint256 invalidParamLength
+    uint256 _invalidParamLength
   ) public {
-    invalidParamLength = bound(invalidParamLength, 1, 256);
-    vm.assume(invalidParamLength != 48);
-    bytes memory fractionalVoteData = new bytes(invalidParamLength);
+    _invalidParamLength = bound(_invalidParamLength, 1, 255);
```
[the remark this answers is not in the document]
davidlaprade: Actually, this number should be much higher. Users could send any crazy large amount of data. I'll update

test/GovernorCountingFractional.t.sol
```diff
@@ -454,33 +499,42 @@ contract GovernorCountingFractionalTest is Test {
...
+    vm.assume(_invalidParamLength != 48);
+
+    // Construct a byte array of arbitrary length and arbitrary values.
+    bytes memory _invalidVoteData = new bytes(_invalidParamLength);
```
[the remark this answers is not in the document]
davidlaprade: Wow yeah, for some reason I didn't think you could fuzz over bytes. This is way better, thanks!

test/GovernorCountingFractional.t.sol
```diff
@@ -599,4 +654,273 @@ contract GovernorCountingFractionalTest is Test {
...
+    address _voterAddr,
+    uint256 _weight,
+    uint256 _votePercentage1,
+    uint256 _votePercentage2,
+    uint256 _votePercentage3,
+    FractionalVoteSplit memory _voteSplit
+  ) public {
+    // Build the vote split.
+    _voteSplit = _randomVoteSplit(_voteSplit);
+    _votePercentage1 = bound(_votePercentage1, 0.0e18, 1.0e18);
```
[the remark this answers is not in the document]
davidlaprade: The setup is confusing, I agree. I had to read a bunch of stuff to refresh myself upon stepping back in here too. I'll add some more comments

test/GovernorCountingFractional.t.sol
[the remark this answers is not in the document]
mds1: Oh whoops, I was thinking the range here had a min of `_weight` and max of `MAX_VOTE_WEIGHT + 1` 🙃 Yes makes sense

test/GovernorCountingFractional.t.sol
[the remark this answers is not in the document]
mds1: Going to mark this resolved in favor of just using a `bytes` fuzzer input as discussed in https://github.com/ScopeLift/flexible-voting/pull/31#discussion_r1088109807

test/GovernorCountingFractional.t.sol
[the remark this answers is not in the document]
mds1: You of course can change the timestamp with `vm.warp`, but yea by default no blocks are mined and no time elapses unless you use a cheatcode to do so

test/GovernorCountingFractional.t.sol
```diff
@@ -599,4 +654,273 @@ contract GovernorCountingFractionalTest is Test {
...
+    assertEq(_firstVote.forVotes + _secondVote.forVotes + _thirdVote.forVotes, _actualForVotes);
+    assertEq(
+      _firstVote.againstVotes + _secondVote.againstVotes + _thirdVote.againstVotes,
+      _actualAgainstVotes
+    );
+    assertEq(
+      _firstVote.abstainVotes + _secondVote.abstainVotes + _thirdVote.abstainVotes,
+      _actualAbstainVotes
+    );
+  }
```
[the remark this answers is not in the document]
davidlaprade: I feel like this should be enough. Given [5000 fuzz runs](https://github.com/ScopeLift/flexible-voting/blob/debfa653e2e18d2421c9aa405db228baee58f8a1/foundry.toml#L11) the odds seem pretty good that we end up with at least one in which all of the weight is cast.

To make sure, though, I'll add a couple non-fuzz tests that specify the percentages so that we certainly will run tests that exercise both full weight and partial weight

src/GovernorCountingFractional.sol
```diff
@@ -98,13 +98,17 @@ abstract contract GovernorCountingFractional is Governor {
         uint256 weight,
         bytes memory voteData
     ) internal virtual override {
-        require(!_proposalVotersHasVoted[proposalId][account], "GovernorCountingFractional: vote already cast");
-        _proposalVotersHasVoted[proposalId][account] = true;
+        require(weight > 0, "GovernorCountingFractional: no weight");
+        if (_proposalVotersWeightCast[proposalId][account] == weight) {
+          revert("GovernorCountingFractional: vote already cast");
+        }
```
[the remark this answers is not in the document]
davidlaprade: > Why not make this also a require for consistency?

It was a require and I removed it in 455cc1f4a2430f96decec6e4952faebbbc2f4185 because I found it confusing. I think this is more self-documenting.

> should we tweak the error string now that you can vote multiple times?

Sure, that's fair

src/GovernorCountingFractional.sol
davidlaprade: Noted. Though I'll be honest that I made all of these changes because I expected to get comments on it if I didn't update the old code to fit the current style guide

src/GovernorCountingFractional.sol
davidlaprade: Thanks, nice catch!

test/GovernorCountingFractional.t.sol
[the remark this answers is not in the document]
davidlaprade: Done in 575f596

test/GovernorCountingFractional.t.sol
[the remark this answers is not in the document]
davidlaprade: Done in 575f596

test/GovernorCountingFractional.t.sol
```diff
@@ -249,35 +265,37 @@ contract GovernorCountingFractionalTest is Test {
...
+      vm.expectEmit(true, false, false, true);
+      emit VoteCast(_voter.addr, _proposalId, _voter.support, _voter.weight, "Yay");
+    }
+
+    vm.prank(_voter.addr);
+    governor.castVoteWithReasonAndParams(_proposalId, _voter.support, "Yay", fractionalizedVotes);
 
-      vm.prank(voter.addr);
-      governor.castVoteWithReasonAndParams(_proposalId, voter.support, "Yay", fractionalizedVotes);
+    assert(governor.hasVoted(_proposalId, _voter.addr));
```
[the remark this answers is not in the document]
davidlaprade: Done in 575f596

test/GovernorCountingFractional.t.sol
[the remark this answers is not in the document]
davidlaprade: Removed in 575f596

test/GovernorCountingFractional.t.sol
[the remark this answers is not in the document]
davidlaprade: Done in 575f596

test/GovernorCountingFractional.t.sol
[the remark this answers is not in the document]
davidlaprade: Done in 575f596

test/GovernorCountingFractional.t.sol
```diff
@@ -599,4 +654,273 @@ contract GovernorCountingFractionalTest is Test {
     assertEq(_againstVotes, _actualAgainstVotes);
     assertEq(_abstainVotes, _actualAbstainVotes);
   }
+
+  function testFuzz_CanCastPartialWeightMultipleTimes(
```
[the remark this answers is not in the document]
davidlaprade: Done in 575f596

test/GovernorCountingFractional.t.sol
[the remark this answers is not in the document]
davidlaprade: Done in 575f596

test/GovernorCountingFractional.t.sol
```diff
@@ -599,4 +654,273 @@ contract GovernorCountingFractionalTest is Test {
...
+  function testFuzz_NominalVotingCannotExceedOverallWeightWithMultipleVotes(
+    uint256[4] memory _weights
+  ) public {
+    Voter memory _voter = _setupNominalVoters(_weights)[0];
+    _mintAndDelegateToVoter(_voter);
+
+    uint256 _proposalId = _createAndSubmitProposal();
+    bytes memory _emptyDataBecauseWereVotingNominally;
+
+    vm.expectEmit(true, false, false, true);
```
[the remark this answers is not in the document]
davidlaprade: Done in 575f596

test/GovernorCountingFractional.t.sol
```diff
@@ -599,4 +654,273 @@ contract GovernorCountingFractionalTest is Test {
...
+      uint128(_voter.weight.mulWadDown(_voteSplit.percentAbstain)) // abstainVotes
+    );
+
+    // Mint, delegate, and propose.
+    _mintAndDelegateToVoter(_voter);
+    uint256 _proposalId = _createAndSubmitProposal();
+
+    if (_isCastingNominallyFirst) {
+      // Vote nominally. It should succeed.
+      vm.expectEmit(true, false, false, true);
```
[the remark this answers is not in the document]
davidlaprade: Done in 575f596

src/GovernorCountingFractional.sol
```diff
@@ -41,7 +41,7 @@ abstract contract GovernorCountingFractional is Governor {
      * @dev See {IGovernor-hasVoted}.
      */
     function hasVoted(uint256 proposalId, address account) public view virtual override returns (bool) {
-        return _proposalVotersHasVoted[proposalId][account];
+        return _proposalVotersWeightCast[proposalId][account] > 0;
```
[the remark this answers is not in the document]
davidlaprade: Done in 575f596

src/GovernorCountingFractional.sol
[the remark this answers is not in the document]
davidlaprade: Error string tweaked in 575f596

src/GovernorCountingFractional.sol
davidlaprade: Done in 575f596

test/GovernorCountingFractional.t.sol
```diff
@@ -249,35 +264,37 @@ contract GovernorCountingFractionalTest is Test {
...
 
-  function _castVotes(Voter[4] memory voters, uint256 _proposalId) internal {
-    Voter memory voter;
-    for (uint8 _i = 0; _i < voters.length; _i++) {
-      voter = voters[_i];
-
-      assert(!governor.hasVoted(_proposalId, voter.addr));
+  function _castVotes(Voter memory _voter, uint256 _proposalId) internal {
+    if (_voter.weight == 0) return;
+    assertEq(governor.hasVoted(_proposalId, _voter.addr), false);
```
mds1, line 269: Nit, you can use `assertFalse(governor.hasVoted(_proposalId, _voter.addr))` instead. Feel free to resolve this comment yourself.

test/GovernorCountingFractional.t.sol
```diff
@@ -249,35 +264,37 @@ contract GovernorCountingFractionalTest is Test {
...
+      vm.expectEmit(true, true, true, true);
+      emit VoteCast(_voter.addr, _proposalId, _voter.support, _voter.weight, "Yay");
+    }
+
+    vm.prank(_voter.addr);
+    governor.castVoteWithReasonAndParams(_proposalId, _voter.support, "Yay", fractionalizedVotes);
 
-      vm.prank(voter.addr);
-      governor.castVoteWithReasonAndParams(_proposalId, voter.support, "Yay", fractionalizedVotes);
+    assertEq(governor.hasVoted(_proposalId, _voter.addr), true);
```
mds1, line 292: Nit, you can use `assertTrue(governor.hasVoted(_proposalId, _voter.addr))` instead. Feel free to resolve this comment yourself.

src/GovernorCountingFractional.sol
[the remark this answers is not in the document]
mds1: Fair enough, lgtm!

src/GovernorCountingFractional.sol
mds1: Personally I don't like the leading underscore convention, so you wouldn't have gotten comments from me 😛 (I find it adds a lot of noise and hurts readability without adding much value, and there's other ways to solve the the same problem anyway)

mds1: Either way, resolving 👌

mds1 (approved):

src/GovernorCountingFractional.sol
apbendi, line 44: Hmm, I think there's a subtle issue here that I'm a little worried about: currently if `hasVoted` returns `true` there is an implicit assumption that they won't vote again and their votes won't change. We're changing that, and I could see it breaking tooling or extension contracts that make that assumption. Not sure what if any action we can take is, though. Hmmm.

src/GovernorCountingFractional.sol
```diff
@@ -98,13 +98,17 @@ abstract contract GovernorCountingFractional is Governor {
         uint256 weight,
         bytes memory voteData
     ) internal virtual override {
-        require(!_proposalVotersHasVoted[proposalId][account], "GovernorCountingFractional: vote already cast");
-        _proposalVotersHasVoted[proposalId][account] = true;
+        require(weight > 0, "GovernorCountingFractional: no weight");
+        if (_proposalVotersWeightCast[proposalId][account] == weight) {
```
apbendi, line 102: Might be better to defensively check `>=`

src/GovernorCountingFractional.sol
```diff
@@ -134,23 +146,28 @@ abstract contract GovernorCountingFractional is Governor {
...
 
-        (uint128 forVotes, uint128 againstVotes, uint128 abstainVotes) = _decodePackedVotes(voteData);
+        (uint128 _forVotes, uint128 _againstVotes, uint128 _abstainVotes) = _decodePackedVotes(voteData);
 
-        require(
-            uint256(forVotes) + againstVotes + abstainVotes <= uint128(weight),
-            "GovernorCountingFractional: votes exceed weight"
-        );
+        uint128 _existingWeight = _proposalVotersWeightCast[proposalId][account];
+        uint256 _newWeight = uint256(_forVotes) + _againstVotes + _abstainVotes + _existingWeight;
```
apbendi, line 158: So, if I understand how this implementation works, the call to castVote submits the marginal votes (i.e. the new votes) to the contract. This works, but I'm trying to think through the implications for the delegate contract, which now needs to track how much weight has or hasn't been submitted, or else look it up from the Governor.

The alternative approach would be to assume the delegate contract always votes with 100% of votes each time, and then have our Governor check that the For/Against/Abstain values for the current vote are less than the For/Against/Abstain values for the previous vote.

Not sure one is better than the other just trying to make sure we've considered the tradeoffs. What are your thoughts?

apbendi (commented):
Looking good @davidlaprade, thanks!

src/GovernorCountingFractional.sol
mds1: Interesting point. The [docs](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/bc6de21fe229520b33d13a4d60bbeb1bc5f91bcd/contracts/governance/IGovernor.sol#L185-L189) for `hasVoted` are (understandly) ambiguous on this point:

> Returns whether `account` has cast a vote on `proposalId`.

So we aren't breaking that definition. But to your point people almost certainly have interpreted this as a binary "all or none", since there was no flex voting until now. I think the options are:
1. Do nothing, since this is technically fine. Confirm this is ok and won't break much by searching for some on/off-chain `hasVoted` usage via sourcegraph
2. Add a new `hasFinishedVoting` that returns true if you've voted with all your weight, to ensure there's a method with the original implied semantics
3. Restore the implied semantics of this method and add a `hasPartiallyVoted` method 

I'm personally ok with any of these, perhaps we pull this out into a follow up issue and ask OZ what they'd prefer, to increases chances of getting this upstreamed?

src/GovernorCountingFractional.sol
mds1: Good callout, I don't think we really discussed and I think this approach was just implied.

Seems the current approach is cheaper because we don't currently track an individual user's split, whereas voting with the new full total amount would require us to track that for each user, so I think you'd end up doing more reading/writing than currently

src/GovernorCountingFractional.sol
davidlaprade: Yeah, it's a good point. I'm also not sure what to do about this. Having this return `false` until all weight is cast seems confusing as well -- maybe even more so.

src/GovernorCountingFractional.sol
davidlaprade: > This works, but I'm trying to think through the implications for the delegate contract, which now needs to track how much weight has or hasn't been submitted, or else look it up from the Governor

Does it? I don't think that's necessarily true. For example, when we remove `CAST_VOTE_WINDOW` in our flex voting AToken in #18 I don't think we'll need to do it. As long as the client/delegate contract knows how many votes any given user is owed for a given proposal, and it does this internal accounting properly, that should be sufficient.

Just to make this concrete, here's an example:
* Alice deposits 25 UNI into a flex voting aToken
* Bob deposits 60 UNI
* Charlie deposits 15 UNI
* so 100 UNI has been deposited in total now
* Dave borrows 80 UNI
* the aToken's UNI balance is now 20 UNI
* a UNI proposal is issued
* Alice expresses her FOR vote to the aToken
* Bob expresses an AGAINST vote to the aToken
* someone calls `aToken.castVote`, rolling these two votes up together and submitting them fractionally to the governor
  * to do this, the aToken asks the governance token what its balance (i.e. voting weight) was at the proposal snapshot, it is 20 UNI
  * the aToken knows that Alice is responsible for 25 of the 100 total UNI deposited, so it casts 5 FOR votes for her (25% of its available 20 UNI voting weight)
  * the aToken knows that Bob is responsible for 60 of the 100 total UNI deposited, so it casts 12 AGAINST votes for him (60% of its available 20 UNI voting weight)
* sometime later, Charlie expresses a FOR vote to the aToken, then calls `aToken.castVote`
  * the aToken doesn't need to lookup its remaining votes, since it knows that Charlie is responsible for 15 of the 100 total UNI deposited, so it casts 3 FOR votes for him (15% of its available 20 UNI voting weight)

src/GovernorCountingFractional.sol
davidlaprade: Curious what @mds1 thinks of this issue

test/GovernorCountingFractional.t.sol
davidlaprade: Thanks. Done in b202e74

test/GovernorCountingFractional.t.sol
davidlaprade: Thanks. Done in b202e74

src/GovernorCountingFractional.sol
davidlaprade: Thanks. Done in b202e74

src/GovernorCountingFractional.sol
mds1: > This works, but I'm trying to think through the implications for the delegate contract, which now needs to track how much weight has or hasn't been submitted, or else look it up from the Governor

Ah I overlooked this and didn't address it. I agree with @davidlaprade here though, I don't think this has downsides for the delegate contract (aToken, in this example). Instead of (from the above example) this flow:

- Alice expresses her FOR vote to the aToken
- Bob expresses an AGAINST vote to the aToken
- someone calls aToken.castVote, rolling these two votes up together and submitting them fractionally to the governor

You can also have the delegate contract designed as follows:

- Alice casts her FOR vote to the aToken, by calling `aToken.castVote()` which forwards her vote to the governor
- Bob expresses an AGAINST vote to the aToken, by calling `aToken.castVote()` which forwards his vote to the governor

In either flow, the aToken contract just needs to have one storage write per vote (whether my cast vote or david's express vote) indicating alice voted. Currently, it does a storage write anyway to save off the expressed vote counts, so you just change what gets stored

davidlaprade: Yup agreed, that would also work

apbendi: @davidlaprade in your example above, when Charlie calls `atToken.castVote`, how does the aToken know to submit only his vote, and not to resbumit Alice and Bob's?

davidlaprade: > how does the aToken know to submit only his vote, and not to resbumit Alice and Bob's

It would have to clear Alice and Bob's votes from storage after submitting them

davidlaprade: Okay, I implemented both approaches so that we can compare:
* [incremental partial voting](https://github.com/ScopeLift/flexible-voting/pull/33) (i.e. what this PR assumed)
* [replacement parial voting](https://github.com/ScopeLift/flexible-voting/pull/34) (i.e. the alternative Ben pointed out)

Grain of salt: the Governor diffs aren't a fair gauge of complexity since they are both based on this PR and this PR already implemented incremental partial voting in the Governor, so that diff for #33 is going to be smaller. The better place to compare complexity is in the client, i.e. the changes to `ATokenFlexVoting`. And there I think you'll see that the two are basically identical in complexity.

With respect to gas costs, the two functions we care about are`expressVote` and `castVote`. Here's how they compare between the two implementations:

| function | implementation | min | avg | median | max |
|----------|----------------|-----|-----|--------|-----|
| expressVote | replacement | 2467            | 36498  | 47776  | 48320  |
| expressVote | incremental | 2489            | 37316  | 47798  | 48342  |
| castVote       | replacement | 12076           | 66037  | 70596  | 108396 |
| castVote       | incremental | 12053           | **49295**  | **53887**  | **68207**  |

That last row is pretty dramatic.

One last data point, I asked @jferas @wildmolasses and @garyghayrat the following question:

> If I told you that a governance system allowed you to cast partial votes, how would you assume that system tallied the votes you cast? More concretely: if you had 20 weight to cast, and you voted 3 times -- first by casting 3 votes, next by casting 7 votes, and third by casting 10 votes -- how much of your voting weight would you assume had been used up in the process?

Everyone assumed that in a scenario like the one described, you would have cast all of your weight with these votes, i.e. the default assumption was that the governor would treat partial votes incrementally.

By no means is this conclusive, but I think it does say something about ease of integration that 5/6 assumed the incremental behavior was how this would work.

davidlaprade: As a note: @apbendi, @mds1, and I just got on a call and discussed this and agreed that the incremental approach is probably best. It's cheaper and doesn't seem like it has a downside for the client.

An idempotent approach (like the replacement approach Ben suggested) would work really well in an environment where you have to worry about lock contention. But that isn't a concern here, since we effectively hold a global lock within the context of each transaction.

src/GovernorCountingFractional.sol
davidlaprade: @rsolari (since I know you discussed this feature with @apbendi) I'm curious if you have any take on this question.

TLDR: how should `hasVoted` work if it's possible for people to cast partial votes?

Some options:
1. it could return true so long as some weight has been cast, even if not all weight has been cast
1. it could return false unless all weight has been cast
1. some combination of (1) and then also add another function like `hasFullyVoted` to capture the behavior that `hasVoted` used to have
1. something else… we’re all ears

What would be best from your (Tally's) point of view

rsolari: Tally doesn't use `hasVoted()` right now, so I don't have a strong opinion from Tally's perspective.

I can imagine a frontend using `hasVoted()` to see whether a user is still able to vote. I could also imagine a subgraph/indexer to using it to count up votes cast. 

For a Flexible Voter, both an indexer and a frontend would need to know how many votes are cast so far. We can't answer that with a boolean, so I don't think option 3 really helps. I'd suggest option 1 because it keeps the semantics from `IGovernor`: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/91e8d0ba3c3beb8a1db31310d8599664e48639ef/contracts/governance/IGovernor.sol#L187

I'd suggest adding a helper e.g. `votesCast(uint256 proposalId, address account)` for getting the votes so far. Tally would just use an indexer to avoid RPC calls, but it's nice to give the option to smart contracts and frontends.

mds1: @davidlaprade mind pulling this out into a separate issue, then we can mark resolved?

src/GovernorCountingFractional.sol
mds1: Probably can resolve this thread now?

mds1 (approved):

src/GovernorCountingFractional.sol
davidlaprade: Awesome, thanks so much for the input @rsolari!

I think I agree that adding a new helper would be the best approach here. Tracking this in https://github.com/ScopeLift/flexible-voting/issues/37

apbendi (approved):

PR #32, opened 2023-01-25
ATokenFlexVoting: Test that borrowers can vote and that votes aren't double counted

davidlaprade:
We need to make sure that liquidity is taken into consideration when casting votes from aTokens. If some of the governance tokens have been borrowed, this should reduce the amount that the aToken can vote with. We don't want to risk double counting votes!

This was a concern of some folks from Aave, per telegram chat, so I've added some additional test assertions to check for it

### DIFF
--- a/test/ATokenFlexVotingFork.t.sol
+++ b/test/ATokenFlexVotingFork.t.sol
@@ -1261,6 +1261,20 @@ contract CastVote is AaveAtokenForkTest {
       if (_vars.supportTypeB == uint8(VoteType.Abstain)) assertApproxEqAbs(_abstainVotes, _expectedVotingWeightB, 1);
     }
     // forgefmt: disable-end
+
+    // The borrower should also be able to submit votes!
+    vm.prank(_vars.borrower);
+    governor.castVoteWithReasonAndParams(
+      _proposalId,
+      uint8(VoteType.For),
+      "Vote from the person that borrowed Gov from Aave",
+      new bytes(0) // Vote nominally so that all of the borrower's weight is used.
+    );
+
+    // The summed votes should not exceed the amount of Gov initially supplied.
+    (_againstVotes, _forVotes, _abstainVotes) = governor.proposalVotes(_proposalId);
+    assertGe(_initGovBalance, _againstVotes + _forVotes + _abstainVotes);
+    assertGe(_vars.voteWeightA + _vars.voteWeightB, _againstVotes + _forVotes + _abstainVotes);
   }
 
   struct VotingWeightIsAbandonedVars {


test/ATokenFlexVotingFork.t.sol
```diff
@@ -1261,6 +1261,20 @@ contract CastVote is AaveAtokenForkTest {
...
+    // The borrower should also be able to submit votes!
+    vm.prank(_vars.borrower);
+    governor.castVoteWithReasonAndParams(
+      _proposalId,
+      uint8(VoteType.For),
+      "Vote from the person that borrowed Gov from Aave",
+      new bytes(0) // Vote nominally so that all of the borrower's weight is used.
+    );
+
+    // The summed votes should not exceed the amount of Gov initially supplied.
```
mds1, line 1274: Are we voting with 100% of weight here? Or is it possible this check always passes because we just happen to be voting with small enough weight? (This is a big test so seemed easier to just ask than try to figure it out 😅)

davidlaprade: Good question!

> Are we voting with 100% of weight here?

Yup, that's right. You can see [here](https://github.com/ScopeLift/flexible-voting/blob/debfa653e2e18d2421c9aa405db228baee58f8a1/test/ATokenFlexVotingFork.t.sol#L1244) that the total votes recorded by the governor after the aToken votes is equal to the [aToken's governance token balance](https://github.com/ScopeLift/flexible-voting/blob/debfa653e2e18d2421c9aa405db228baee58f8a1/test/ATokenFlexVotingFork.t.sol#L1223). So all of its weight is cast.

mds1: Got it! Where does the off by one error come from? Guessing for/against/abstain votes are all computed with the `x * y / z` format which always rounds down, and we can replace the final calculation with subtraction instead?

davidlaprade: > Where does the off by one error come from? Guessing for/against/abstain votes are all computed with the x * y / z format which always rounds down

Yup, correct.

> we can replace the final calculation with subtraction instead?

Not sure what you mean here, sorry

mds1: For example (pseudocode below), change from this:

```solidity
forVotes = voter.weight * percentFor / votePercentage
againstVotes = voter.weight * percentAgainst / votePercentage
abstainVotes = voter.weight * percentAbstain / votePercentage
```

to this:

```solidity
forVotes = voter.weight * percentFor / votePercentage
againstVotes = voter.weight * percentAgainst / votePercentage
abstainVotes = voter.weight - forVotes - againstVotes
```

(This may not be exactly correct cause of the `votePercentage` scaling but I think you'll get the idea)

davidlaprade: Oh oh, I see what you mean. Thanks for clarifying!

Unfortunately, I don't think we can't do that to fix the loss of precision here, since the rounding occurs when the aToken casts the rolled up vote to the governor. It specifically happens [here](https://github.com/ScopeLift/flexible-voting/blob/debfa653e2e18d2421c9aa405db228baee58f8a1/src/ATokenFlexVoting.sol#L250-L263). We can't use subtraction because the aToken might not be casting with its entire weight, e.g. one of the aToken holders might not have expressed a voting preference on the proposal.

This is going to become even more important once we merge #31, in which partial weight votes could very well be the norm. Subtraction won't be an option.

mds1: Ah interesting, hmm ok think I'm good with leaving this as-is. Makes sense. Going to leave unresolved just so @apbendi sees this when reviewing

apbendi (approved):
Looks good to me!

mds1 (approved):

PR #33, opened 2023-01-30
Support incremental voting on ATokenFlexVoting

davidlaprade:
Compare to the approach in #34. The idea here is that partial votes to `GovernorCountingFractional` will increment the amounts previously cast.

To give a concrete example:
* you have 100 weight to vote with
* the first time you vote, you cast 10 FOR, 20 AGAINST, 0 ABSTAIN
* the governor records 30 votes for you
* the next time you vote, you cast 30 FOR, 5 AGAINST, 10 ABSTAIN
* the governor now records 75 votes for you -- these new votes were added to your previous totals
* finally, you attempt to cast the following votes: 0 FOR, 25 AGAINST, 10 ABSTAIN
* the governor reverts, since if these votes were added to your previous votes you would have voted with a weight of 110, which is more than you have

Here's the gas report run with `forge test --gas-report` (I highlighted the two functions we care most about):

image

This was done to help make a decision on the discussion [here](https://github.com/ScopeLift/flexible-voting/pull/31#discussion_r1089299026).

I didn't fix all of the tests on this branch (though there are only a couple more that are failing). I just verified that the main feature test that I wrote passed: `test_VotesCanBeCastIncrementally`. If we decide to go with this approach, more tests should be written.

This PR closes https://github.com/ScopeLift/flexible-voting/issues/18

### DIFF
--- a/src/ATokenFlexVoting.sol
+++ b/src/ATokenFlexVoting.sol
@@ -192,6 +192,10 @@ contract ATokenFlexVoting is AToken {
   /// @param proposalId The ID of the proposal which the Pool will now vote on.
   function castVote(uint256 proposalId) external {
     ProposalVote storage _proposalVote = proposalVotes[proposalId];
+    require(
+      _proposalVote.forVotes + _proposalVote.againstVotes + _proposalVote.abstainVotes > 0,
+      "no votes expressed"
+    );
     uint256 _proposalSnapshotBlockNumber = GOVERNOR.proposalSnapshot(proposalId);
 
     // Use the snapshot of total raw balances to determine total voting weight.
--- a/src/ATokenFlexVoting.sol
+++ b/src/ATokenFlexVoting.sol
@@ -23,7 +23,7 @@ import {IVotingToken} from "src/interfaces/IVotingToken.sol";
 /// proposals. When they do so, this extension records that preference with weight proportional to
 /// the users's AToken balance at the proposal snapshot.
 ///
-/// At any point after voting preferrences have been expressed, the AToken's public `castVote`
+/// At any point after voting preferences have been expressed, the AToken's public `castVote`
 /// function may be called to roll up all internal voting records into a single delegated vote to
 /// the Governor contract -- a vote which specifies the exact For/Abstain/Against totals expressed
 /// by AToken holders. Votes can be rolled up and cast in this manner multiple times for a given
@@ -189,7 +189,7 @@ contract ATokenFlexVoting is AToken {
 
   /// @notice Causes this contract to cast a vote to the Governor for all of the
   /// accumulated votes expressed by users. Uses the sum of all raw (unrebased) balances
-  /// to proportion and split its voting weight. Can be called by anyone. Can be called
+  /// to proportionally split its voting weight. Can be called by anyone. Can be called
   /// multiple times during the lifecycle of a given proposal.
   /// @param proposalId The ID of the proposal which the Pool will now vote on.
   function castVote(uint256 proposalId) external {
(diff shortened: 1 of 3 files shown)


test/ATokenFlexVotingFork.t.sol
```diff
@@ -1658,6 +1581,62 @@ contract CastVote is AaveAtokenForkTest {
...
+    aToken.castVote(_proposalId);
+
+    (_againstVotes, _forVotes, _abstainVotes) = governor.proposalVotes(_proposalId);
+
+    if (_supportTypeA == uint8(VoteType.For)) assertEq(_forVotes, _weightA);
+    if (_supportTypeA == uint8(VoteType.Against)) assertEq(_againstVotes, _weightA);
+    if (_supportTypeA == uint8(VoteType.Abstain)) assertEq(_abstainVotes, _weightA);
+    if (_supportTypeB == uint8(VoteType.For)) assertEq(_forVotes, _weightB);
+    if (_supportTypeB == uint8(VoteType.Against)) assertEq(_againstVotes, _weightB);
+    if (_supportTypeB == uint8(VoteType.Abstain)) assertEq(_abstainVotes, _weightB);
```
mds1, line 1638: There will always be an unused support type since there's only two voters here, should we have if statements + assertions to ensure that third type has zero votes?

src/ATokenFlexVoting.sol
```diff
@@ -23,9 +23,11 @@ import {IVotingToken} from "src/interfaces/IVotingToken.sol";
 /// proposals. When they do so, this extension records that preference with weight proportional to
 /// the users's AToken balance at the proposal snapshot.
 ///
-/// When the proposal deadline nears, the AToken's public `castVote` function is called to roll up
-/// all internal voting records into a single delegated vote to the Governor contract -- a vote
-/// which specifies the exact For/Abstain/Against totals expressed by AToken holders.
+/// At any point after voting preferrences have been expressed, the AToken's public `castVote`
```
mds1, line 26:
```suggestion
/// At any point after voting preferences have been expressed, the AToken's public `castVote`
```

src/ATokenFlexVoting.sol
```diff
@@ -214,32 +187,35 @@ contract ATokenFlexVoting is AToken {
...
 
-  /// @notice Causes this contract to cast a vote to the Governor for all the
-  /// tokens it currently holds. Uses the sum of all depositor voting
-  /// expressions to decide how to split its voting weight. Can be called by
-  /// anyone, but _must_ be called within `CAST_VOTE_WINDOW` blocks before the
-  /// proposal deadline. We don't bother to check if the vote has already been
-  /// cast -- GovernorCountingFractional will revert if it has.
+  /// @notice Causes this contract to cast a vote to the Governor for all of the
+  /// accumulated votes expressed by users. Uses the sum of all raw (unrebased) balances
+  /// to proportion and split its voting weight. Can be called by anyone. Can be called
```
mds1, line 192:
```suggestion
  /// to proportionally split its voting weight. Can be called by anyone. Can be called
```

src/ATokenFlexVoting.sol
```diff
@@ -214,32 +187,35 @@ contract ATokenFlexVoting is AToken {
...
-    // We cannot use the proposalVote numbers alone, since some people with
-    // balances at the snapshot might not have expressed votes. We don't want to
-    // make it possible for aToken holders to *increase* their voting power when
-    // other people don't express their votes. That'd be a terrible incentive.
+    // We use the snapshot of total raw balances to determine the weight with
+    // which to vote. We do this for two reasons:
+    //   (1) We cannot use the proposalVote numbers alone, since some people with
+    //       balances at the snapshot might never express their preferences. If a
+    //       large holder never expressed a preference, but this contract nevertheless
+    //       cast votes to the governor with all of its weight, then other users would
```
mds1, line 208: Small clarification since it's not guaranteed that your voting weight would have increased, since it's also dependent on how many tokens are borrowed (which isn't mentioned in this comment block, but maybe should be)
```suggestion
    //       cast votes to the governor with all of its weight, then other users may
```

mds1 (commented):
Looks great! All small comments that you can self-resolve

test/ATokenFlexVotingFork.t.sol
davidlaprade: Good point. Clarified the assertions in 3976218

mds1 (approved):

apbendi (approved):
Looks good to me!

PR #34, opened 2023-01-31
[wip] Replacement voting spike

davidlaprade:
Compare to the approach in #33. The idea here is that partial votes to `GovernorCountingFractional` will "replace" one another -- so long as the new voting weight of each support type is >= its existing value.

To give a concrete example:
* you have 100 weight to vote with
* the first time you vote, you cast 10 FOR, 20 AGAINST, 0 ABSTAIN
* the governor records 30 votes for you
* the next time you vote, you cast 40 FOR, 25 AGAINST, 10 ABSTAIN
* the governor now records 75 votes for you -- these new votes replaced your previous totals
* finally, you attempt to cast the following votes: 70 FOR, 20 AGAINST, 10 ABSTAIN
* the governor reverts, since your new AGAINST total is lower than your previously cast total

Here's the gas report run with `forge test --gas-report` (I highlighted the two functions we care most about):

image

This was done to help make a decision on the discussion [here](https://github.com/ScopeLift/flexible-voting/pull/31#discussion_r1089299026).

I didn't fix all of the tests on this branch (though there are only a couple more that are failing). I just verified that the main feature test that I wrote passed: `test_VotesCanBeCastWithReplacement`. If we decide to go with this approach, more tests should be written.

### DIFF
--- a/src/GovernorCountingFractional.sol
+++ b/src/GovernorCountingFractional.sol
@@ -26,17 +26,22 @@ abstract contract GovernorCountingFractional is Governor {
         uint128 abstainVotes;
     }
 
+    function _totalProposalVoteWeights(ProposalVote memory _vote) internal view returns(uint256) {
+      return uint256(_vote.againstVotes) + _vote.forVotes + _vote.abstainVotes;
+    }
+
     /**
      * @dev Mapping from proposal ID to vote tallies for that proposal.
      */
     mapping(uint256 => ProposalVote) private _proposalVotes;
 
     /**
-     * @dev Mapping from proposal ID and address to the weight the address
-     * has cast on that proposal, e.g. _proposalVotersWeightCast[42][0xBEEF]
-     * would tell you the number of votes that 0xBEEF has cast on proposal 42.
+     * @dev Mapping from proposal ID and address to the votes the address
+     * has cast on that proposal, e.g. _proposalVotesCast[42][0xBEEF]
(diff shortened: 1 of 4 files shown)


src/GovernorCountingFractional.sol
```diff
@@ -155,31 +183,45 @@ abstract contract GovernorCountingFractional is Governor {
...
-        _proposalVote = ProposalVote(
-            _proposalVote.againstVotes + _againstVotes,
-            _proposalVote.forVotes + _forVotes,
-            _proposalVote.abstainVotes + _abstainVotes
+        require(_storedVote.againstVotes <= _newVote.againstVotes, "cannot undo past votes");
+        require(_storedVote.forVotes <= _newVote.forVotes, "cannot undo past votes");
+        require(_storedVote.abstainVotes <= _newVote.abstainVotes, "cannot undo past votes");
+
+        // How many votes are we adding here?
+        ProposalVote memory _voteDelta;
```
apbendi, line 203: Why do we calculate the delta instead of just overwriting?

davidlaprade: Good question!

Because the previous votes have already been added to the overall voting tallies. To make sure that we don't double count votes, we only increment the overall tallies by the delta between what was cast before and what is being cast now.

davidlaprade:
Closing out in favor of #33 per discussion [here](https://github.com/ScopeLift/flexible-voting/pull/31#discussion_r1092460706)

PR #40, opened 2023-03-13
M-01 Inconsistent Ordering of Votes

davidlaprade:
In GovernorCountingFractional.sol , the ProposalVote struct is defined in the following order: againstVotes , forVotes , abstainVotes . This order aligns with the standard usage in OpenZeppelin's contracts library as well as in Compound's Bravo.

However, the voteData bytes parameter received by the _countVote function requires a different order: forVotes , againstVotes , abstainVotes . While the code currently handles the decoding of this bytes parameter correctly, this inconsistency in the ordering can be problematic for projects that are integrating with the GovernorCountingFractional
contract.

### DIFF
--- a/src/ATokenFlexVoting.sol
+++ b/src/ATokenFlexVoting.sol
@@ -246,7 +246,7 @@ contract ATokenFlexVoting is AToken {
     delete proposalVotes[proposalId];
 
     bytes memory fractionalizedVotes =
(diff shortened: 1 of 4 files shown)


apbendi (approved):
LGTM!

mds1 (approved):
Thanks!

PR #45, opened 2023-03-15
N-01 OpenZeppelin Imports in Incorrect Format

davidlaprade:
[OZ recommends]( https://docs.openzeppelin.com/contracts/4.x/) that contracts import its code using the node style so that they have out-of-the box compatibility with [hardhat](https://github.com/NomicFoundation/hardhat).

Note that CI will fail because of a formatting error. I've left the formatting change out of this PR to (a) keep it clean and specific to the issue above and (b) because that formatting error has already been fixed in #40 which will merge before this.

I manually tested that this can now be imported into a hardhat project without issue. Steps to do this were:
1. `mkdir` + `cd` into a new directory
1. `npm init`
3. `npx hardhat`
4. `npm install @openzeppelin/contracts`
5. `npm install ScopeLift/flexible-voting#oz-audit-fixes-N01-hardhat-friendly-imports --save`
6. copy [this file](https://github.com/ScopeLift/flexible-voting/blob/master/test/FractionalGovernor.sol) into `/contracts`
7. modify the file so that imports look like:
    ```solidity
      import "flexible-voting/src/GovernorCountingFractional.sol";
      import "@openzeppelin/contracts/governance/extensions/GovernorVotes.sol";
    ```
8. `npx hardhat compile`

image

### DIFF
--- a/foundry.toml
+++ b/foundry.toml
@@ -4,7 +4,10 @@
   # auto-detect solc versions.
   optimizer = true
   optimizer_runs = 10_000_000
-  remappings = ["openzeppelin-contracts/=lib/openzeppelin-contracts/contracts"]
+  remappings = [
+    "@openzeppelin/contracts/=lib/openzeppelin-contracts/contracts",
+    "openzeppelin-contracts/=lib/openzeppelin-contracts/contracts",
+  ]
   verbosity = 3
 
 [profile.ci]
--- a/foundry.toml
+++ b/foundry.toml
@@ -4,10 +4,7 @@
   # auto-detect solc versions.
   optimizer = true
   optimizer_runs = 10_000_000
-  remappings = [
-    "@openzeppelin/contracts/=lib/openzeppelin-contracts/contracts",
-    "openzeppelin-contracts/=lib/openzeppelin-contracts/contracts",
-  ]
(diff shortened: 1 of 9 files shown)


apbendi (approved):
Did not test with Hardhat; have we confirmed this works out of the box with it now? Otherwise looks good to me and works fine in itself.

foundry.toml
```diff
@@ -4,7 +4,10 @@
   # auto-detect solc versions.
   optimizer = true
   optimizer_runs = 10_000_000
-  remappings = ["openzeppelin-contracts/=lib/openzeppelin-contracts/contracts"]
+  remappings = [
+    "@openzeppelin/contracts/=lib/openzeppelin-contracts/contracts",
+    "openzeppelin-contracts/=lib/openzeppelin-contracts/contracts",
```
mds1, line 9: I think it'd be better to change all imports to the `@openzeppelin` style instead of just a subset and having two remappings?

davidlaprade: Fair. Done in [c046e6d](https://github.com/ScopeLift/flexible-voting/pull/45/commits/c046e6df1faaca603e57518fdf9bc7297b911498)

mds1 (approved):

PR #46, opened 2023-05-15
Add Compound V3 client contract

davidlaprade:
This PR adds a flex voting client contract that is compatible with Compound V3. This was developed through support from a grant from Compound [here](https://questbook.app/dashboard/?grantId=0xeb047900b28a9f90f3c0e65768b23e7542a65163&chainId=10&proposalId=0x21b&isRenderingProposalBody=true)

Currently this PR depends upon [a fork of Compound V3](https://github.com/davidlaprade/comet/) (split from main on May 15, 2023). The [only change](https://github.com/davidlaprade/comet/commit/2921780a8cfe94f2e11cf91ebab1301daf4a76d7) was to make `Comet.updateBasePrincipal` virtual so that we could add checkpointing in the client. My intention is to upstream this change after we're aligned on the architecture of the present PR and it is approved/merged.

This PR also abstracts the flex voting client logic into a separate abstract contract called `FlexVotingClient` that can be re-used for client implementations in the future. Doing so allows us to greatly simplify the `ATokenFlexVoting` implementation of previous PRs, e.g. #11.

Note that this PR leaves a few things unfinished, e.g.:
* more thorough testing should be done (see TODO's at the bottom of the test file)
* natspec comments (both for `FlexVotingClient` and for `CometFlexVoting`)
* the Comet dependency should be updated to https://github.com/compound-finance/comet once the change is merged

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,32 +39,33 @@ jobs:
 
       - name: Run tests
         run: forge test
-  #
-  # fmt:
-  #   runs-on: ubuntu-latest
-  #   steps:
-  #     - uses: actions/checkout@v3
-  #
-  #     - name: Install Foundry
-  #       uses: foundry-rs/foundry-toolchain@v1
-  #       with:
-  #         version: nightly
-  #
-  #     - name: Install scopelint
-  #       uses: engineerd/configurator@v0.0.8
-  #       with:
-  #         name: scopelint
-  #         repo: ScopeLift/scopelint
-  #         fromGitHubReleases: true
-  #         version: latest
-  #         pathInArchive: scopelint-x86_64-linux/scopelint
-  #         urlTemplate: https://github.com/ScopeLift/scopelint/releases/download/{{version}}/scopelint-x86_64-linux.tar.xz
-  #         token: ${{ secrets.GITHUB_TOKEN }}
-  #
-  #     - name: Check formatting
-  #       run: |
-  #         scopelint --version
-  #         scopelint check
+
+  fmt:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Install Foundry
+        uses: foundry-rs/foundry-toolchain@v1
+        with:
+          version: nightly
+
+      - name: Install scopelint
+        uses: engineerd/configurator@v0.0.8
+        with:
+          name: scopelint
+          repo: ScopeLift/scopelint
+          fromGitHubReleases: true
+          version: latest
+          pathInArchive: scopelint-x86_64-linux/scopelint
+          urlTemplate: https://github.com/ScopeLift/scopelint/releases/download/{{version}}/scopelint-x86_64-linux.tar.xz
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Check formatting
+        run: forge fmt --check
+        # run: |
+        #   scopelint --version
+        #   scopelint check
 
   # TODO figure out why this is failing and uncomment
   # slither-analyze:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,7 @@ jobs:
       - name: Build contracts
         run: |
           forge --version
-          forge build --sizes
+          forge build
   test:
     runs-on: ubuntu-latest
     steps:
@@ -38,32 +38,32 @@ jobs:
 
       - name: Run tests
         run: forge test
-
-  fmt:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v3
-
-      - name: Install Foundry
-        uses: foundry-rs/foundry-toolchain@v1
-        with:
-          version: nightly
-
-      - name: Install scopelint
-        uses: engineerd/configurator@v0.0.8
-        with:
-          name: scopelint
-          repo: ScopeLift/scopelint
-          fromGitHubReleases: true
-          version: latest
-          pathInArchive: scopelint-x86_64-linux/scopelint
-          urlTemplate: https://github.com/ScopeLift/scopelint/releases/download/{{version}}/scopelint-x86_64-linux.tar.xz
-          token: ${{ secrets.GITHUB_TOKEN }}
-
-      - name: Check formatting
-        run: |
-          scopelint --version
-          scopelint check
+  #
+  # fmt:
+  #   runs-on: ubuntu-latest
+  #   steps:
+  #     - uses: actions/checkout@v3
+  #
+  #     - name: Install Foundry
+  #       uses: foundry-rs/foundry-toolchain@v1
+  #       with:
+  #         version: nightly
+  #
+  #     - name: Install scopelint
+  #       uses: engineerd/configurator@v0.0.8
+  #       with:
+  #         name: scopelint
+  #         repo: ScopeLift/scopelint
+  #         fromGitHubReleases: true
+  #         version: latest
+  #         pathInArchive: scopelint-x86_64-linux/scopelint
+  #         urlTemplate: https://github.com/ScopeLift/scopelint/releases/download/{{version}}/scopelint-x86_64-linux.tar.xz
+  #         token: ${{ secrets.GITHUB_TOKEN }}
+  #
+  #     - name: Check formatting
+  #       run: |
+  #         scopelint --version
+  #         scopelint check
 
   # TODO figure out why this is failing and uncomment
   # slither-analyze:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,7 +37,7 @@ jobs:
           version: nightly
 
       - name: Run tests
-        run: forge test
+        run: forge test -vvvv
   #
   # fmt:
   #   runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,16 +50,16 @@ jobs:
         with:
           version: nightly
 
-      - name: Install scopelint
-        uses: engineerd/configurator@v0.0.8
-        with:
-          name: scopelint
-          repo: ScopeLift/scopelint
-          fromGitHubReleases: true
-          version: latest
-          pathInArchive: scopelint-x86_64-linux/scopelint
-          urlTemplate: https://github.com/ScopeLift/scopelint/releases/download/{{version}}/scopelint-x86_64-linux.tar.xz
-          token: ${{ secrets.GITHUB_TOKEN }}
+      # - name: Install scopelint
+      #   uses: engineerd/configurator@v0.0.8
+      #   with:
+      #     name: scopelint
+      #     repo: ScopeLift/scopelint
+      #     fromGitHubReleases: true
+      #     version: latest
+      #     pathInArchive: scopelint-x86_64-linux/scopelint
+      #     urlTemplate: https://github.com/ScopeLift/scopelint/releases/download/{{version}}/scopelint-x86_64-linux.tar.xz
+      #     token: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Check formatting
         run: forge fmt --check
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,6 +9,7 @@ on:
 
 env:
   FOUNDRY_PROFILE: ci
+  MAINNET_RPC_URL: ${{ secrets.MAINNET_RPC_URL }}
   OPTIMISM_RPC_URL: ${{ secrets.OPTIMISM_RPC_URL }}
 
 jobs:
@@ -37,7 +38,7 @@ jobs:
           version: nightly
 
       - name: Run tests
-        run: forge test -vvvv
+        run: forge test
   #
   # fmt:
   #   runs-on: ubuntu-latest
--- a/.gitmodules
+++ b/.gitmodules
@@ -13,4 +13,4 @@
 	branch = v4.8.0
 [submodule "lib/comet"]
 	path = lib/comet
-	url = https://github.com/compound-finance/comet
+	url = https://github.com/davidlaprade/comet
--- a/.gitmodules
+++ b/.gitmodules
@@ -11,3 +11,6 @@
 	path = lib/openzeppelin-contracts
 	url = https://github.com/openzeppelin/openzeppelin-contracts
 	branch = v4.8.0
+[submodule "lib/comet"]
+	path = lib/comet
+	url = https://github.com/compound-finance/comet
--- a/src/ATokenFlexVoting.sol
+++ b/src/ATokenFlexVoting.sol
@@ -13,6 +13,8 @@ import {SafeCast} from "@openzeppelin/contracts/utils/math/SafeCast.sol";
 import {Checkpoints} from "@openzeppelin/contracts/utils/Checkpoints.sol";
 import {IFractionalGovernor} from "src/interfaces/IFractionalGovernor.sol";
 import {IVotingToken} from "src/interfaces/IVotingToken.sol";
+
+import {FlexVotingClient} from "src/FlexVotingClient.sol";
 // forgefmt: disable-end
 
 /// @notice This is an extension of Aave V3's AToken contract which makes it possible for AToken
@@ -39,46 +41,13 @@ import {IVotingToken} from "src/interfaces/IVotingToken.sol";
 /// The original AToken that this contract extends is viewable here:
 ///
 ///   https://github.com/aave/aave-v3-core/blob/c38c6276/contracts/protocol/tokenization/AToken.sol
-contract ATokenFlexVoting is AToken {
-  using SafeCast for uint256;
+contract ATokenFlexVoting is AToken, FlexVotingClient {
   using Checkpoints for Checkpoints.History;
-
-  /// @notice The voting options corresponding to those used in the Governor.
-  enum VoteType {
-    Against,
-    For,
-    Abstain
-  }
-
-  /// @notice Data structure to store vote preferences expressed by depositors.
-  struct ProposalVote {
-    uint128 againstVotes;
-    uint128 forVotes;
-    uint128 abstainVotes;
-  }
-
-  /// @notice Map proposalId to an address to whether they have voted on this proposal.
-  mapping(uint256 => mapping(address => bool)) private proposalVotersHasVoted;
-
-  /// @notice Map proposalId to vote totals expressed on this proposal.
-  mapping(uint256 => ProposalVote) public proposalVotes;
-
-  /// @notice The governor contract associated with this governance token. It
-  /// must be one that supports fractional voting, e.g. GovernorCountingFractional.
-  IFractionalGovernor public immutable GOVERNOR;
-
-  /// @notice Mapping from address to stored (not rebased) balance checkpoint history.
-  mapping(address => Checkpoints.History) private balanceCheckpoints;
-
-  /// @notice History of total stored (not rebased) balances.
-  Checkpoints.History private totalDepositCheckpoints;
 
   /// @dev Constructor.
   /// @param _pool The address of the Pool contract
   /// @param _governor The address of the flex-voting-compatible governance contract.
-  constructor(IPool _pool, address _governor) AToken(_pool) {
-    GOVERNOR = IFractionalGovernor(_governor);
-  }
+  constructor(IPool _pool, address _governor) AToken(_pool) FlexVotingClient(_governor) {}
 
   // forgefmt: disable-start
   //===========================================================================
@@ -109,7 +78,7 @@ contract ATokenFlexVoting is AToken {
       params
     );
 
-    selfDelegate();
+    FlexVotingClient.selfDelegate();
   }
 
   /// Note: this has been modified from Aave v3's MintableIncentivizedERC20 to
@@ -118,8 +87,10 @@ contract ATokenFlexVoting is AToken {
   /// @inheritdoc MintableIncentivizedERC20
   function _burn(address account, uint128 amount) internal override {
     MintableIncentivizedERC20._burn(account, amount);
-    _checkpointRawBalanceOf(account);
-    totalDepositCheckpoints.push(totalDepositCheckpoints.latest() - amount);
+    FlexVotingClient._checkpointRawBalanceOf(account);
+    FlexVotingClient.totalDepositCheckpoints.push(
+      FlexVotingClient.totalDepositCheckpoints.latest() - amount
+    );
   }
 
   /// Note: this has been modified from Aave v3's MintableIncentivizedERC20 to
@@ -128,8 +99,10 @@ contract ATokenFlexVoting is AToken {
   /// @inheritdoc MintableIncentivizedERC20
   function _mint(address account, uint128 amount) internal override {
     MintableIncentivizedERC20._mint(account, amount);
-    _checkpointRawBalanceOf(account);
-    totalDepositCheckpoints.push(totalDepositCheckpoints.latest() + amount);
+    FlexVotingClient._checkpointRawBalanceOf(account);
+    FlexVotingClient.totalDepositCheckpoints.push(
+      FlexVotingClient.totalDepositCheckpoints.latest() + amount
+    );
   }
 
   /// @dev This has been modified from Aave v3's AToken contract to checkpoint raw balances
@@ -145,136 +118,16 @@ contract ATokenFlexVoting is AToken {
     bool validate
   ) internal virtual override {
     AToken._transfer(from, to, amount, validate);
-    _checkpointRawBalanceOf(from);
-    _checkpointRawBalanceOf(to);
+    FlexVotingClient._checkpointRawBalanceOf(from);
+    FlexVotingClient._checkpointRawBalanceOf(to);
   }
   //===========================================================================
   // END: Aave overrides
   //===========================================================================
   // forgefmt: disable-end
 
-  // Self-delegation cannot be done in the constructor because the aToken is
-  // just a proxy -- it won't share an address with the implementation (i.e.
-  // this code). Instead we do it at the end of `initialize`. But even that won't
-  // handle already-initialized aTokens. For those, we'll need to self-delegate
-  // during the upgrade process. More details in these issues:
-  // https://github.com/aave/aave-v3-core/pull/774
-  // https://github.com/ScopeLift/flexible-voting/issues/16
-  function selfDelegate() public {
-    IVotingToken(GOVERNOR.token()).delegate(address(this));
-  }
-
-  /// @notice Allow a depositor to express their voting preference for a given
-  /// proposal. Their preference is recorded internally but not moved to the
-  /// Governor until `castVote` is called.
-  /// @param proposalId The proposalId in the associated Governor
-  /// @param support The depositor's vote preferences in accordance with the `VoteType` enum.
-  function expressVote(uint256 proposalId, uint8 support) external {
-    uint256 weight = getPastStoredBalance(msg.sender, GOVERNOR.proposalSnapshot(proposalId));
-    require(weight > 0, "no weight");
-
-    require(!proposalVotersHasVoted[proposalId][msg.sender], "already voted");
-    proposalVotersHasVoted[proposalId][msg.sender] = true;
-
-    if (support == uint8(VoteType.Against)) {
-      proposalVotes[proposalId].againstVotes += SafeCast.toUint128(weight);
-    } else if (support == uint8(VoteType.For)) {
-      proposalVotes[proposalId].forVotes += SafeCast.toUint128(weight);
-    } else if (support == uint8(VoteType.Abstain)) {
-      proposalVotes[proposalId].abstainVotes += SafeCast.toUint128(weight);
-    } else {
-      revert("invalid support value, must be included in VoteType enum");
-    }
-  }
-
-  /// @notice Causes this contract to cast a vote to the Governor for all of the
-  /// accumulated votes expressed by users. Uses the sum of all raw (unrebased) balances
-  /// to proportionally split its voting weight. Can be called by anyone. Can be called
-  /// multiple times during the lifecycle of a given proposal.
-  /// @param proposalId The ID of the proposal which the Pool will now vote on.
-  function castVote(uint256 proposalId) external {
-    ProposalVote storage _proposalVote = proposalVotes[proposalId];
-    require(
-      _proposalVote.forVotes + _proposalVote.againstVotes + _proposalVote.abstainVotes > 0,
-      "no votes expressed"
-    );
-    uint256 _proposalSnapshotBlockNumber = GOVERNOR.proposalSnapshot(proposalId);
-
-    // We use the snapshot of total raw balances to determine the weight with
-    // which to vote. We do this for two reasons:
-    //   (1) We cannot use the proposalVote numbers alone, since some people with
-    //       balances at the snapshot might never express their preferences. If a
-    //       large holder never expressed a preference, but this contract nevertheless
-    //       cast votes to the governor with all of its weight, then other users may
-    //       effectively have *increased* their voting weight because someone else
-    //       didn't participate, which creates all kinds of bad incentives.
-    //   (2) Other people might have already expressed their preferences on this
-    //       proposal and had those preferences submitted to the governor by an
-    //       earlier call to this function. The weight of those preferences
-    //       should still be taken into consideration when determining how much
-    //       weight to vote with this time.
-    // Using the total raw balance to proportion votes in this way means that in
-    // many circumstances this function will not cast votes with all of its
-    // weight.
-    uint256 _totalRawBalanceAtSnapshot = getPastTotalBalance(_proposalSnapshotBlockNumber);
-
-    // We need 256 bits because of the multiplication we're about to do.
-    uint256 _votingWeightAtSnapshot = IVotingToken(address(_underlyingAsset)).getPastVotes(
-      address(this), _proposalSnapshotBlockNumber
-    );
-
-    //      forVotesRaw          forVoteWeight
-    // --------------------- = ------------------
-    //     totalRawBalance      totalVoteWeight
-    //
-    // forVoteWeight = forVotesRaw * totalVoteWeight / totalRawBalance
-    uint128 _forVotesToCast = SafeCast.toUint128(
-      (_votingWeightAtSnapshot * _proposalVote.forVotes) / _totalRawBalanceAtSnapshot
-    );
-    uint128 _againstVotesToCast = SafeCast.toUint128(
-      (_votingWeightAtSnapshot * _proposalVote.againstVotes) / _totalRawBalanceAtSnapshot
-    );
-    uint128 _abstainVotesToCast = SafeCast.toUint128(
-      (_votingWeightAtSnapshot * _proposalVote.abstainVotes) / _totalRawBalanceAtSnapshot
-    );
-
-    // This param is ignored by the governor when voting with fractional
-    // weights. It makes no difference what vote type this is.
-    uint8 unusedSupportParam = uint8(VoteType.Abstain);
-
-    // Clear the stored votes so that we don't double-cast them.
-    delete proposalVotes[proposalId];
-
-    bytes memory fractionalizedVotes =
-      abi.encodePacked(_againstVotesToCast, _forVotesToCast, _abstainVotesToCast);
-    GOVERNOR.castVoteWithReasonAndParams(
-      proposalId,
-      unusedSupportParam,
-      "rolled-up vote from aToken holders", // Reason string.
-      fractionalizedVotes
-    );
-  }
-
   /// @notice Returns the _user's current balance in storage.
-  function _rawBalanceOf(address _user) internal view returns (uint256) {
+  function _rawBalanceOf(address _user) internal view override returns (uint256) {
     return _userState[_user].balance;
-  }
-
-  /// @notice Checkpoints the _user's current raw balance.
-  function _checkpointRawBalanceOf(address _user) internal {
-    balanceCheckpoints[_user].push(_rawBalanceOf(_user));
-  }
-
-  /// @notice Returns the _user's balance in storage at the _blockNumber.
-  /// @param _user The account that's historical balance will be looked up.
-  /// @param _blockNumber The block at which to lookup the _user's balance.
-  function getPastStoredBalance(address _user, uint256 _blockNumber) public view returns (uint256) {
-    return balanceCheckpoints[_user].getAtProbablyRecentBlock(_blockNumber);
-  }
-
-  /// @notice Returns the total stored balance of all users at _blockNumber.
-  /// @param _blockNumber The block at which to lookup the total stored balance.
-  function getPastTotalBalance(uint256 _blockNumber) public view returns (uint256) {
-    return totalDepositCheckpoints.getAtProbablyRecentBlock(_blockNumber);
   }
 }
--- a/src/ATokenFlexVoting.sol
+++ b/src/ATokenFlexVoting.sol
@@ -88,8 +88,8 @@ contract ATokenFlexVoting is AToken, FlexVotingClient {
   function _burn(address account, uint128 amount) internal override {
     MintableIncentivizedERC20._burn(account, amount);
     FlexVotingClient._checkpointRawBalanceOf(account);
-    FlexVotingClient.totalDepositCheckpoints.push(
-      FlexVotingClient.totalDepositCheckpoints.latest() - amount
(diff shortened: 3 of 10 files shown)


.github/workflows/ci.yml
```diff
@@ -25,7 +26,7 @@ jobs:
       - name: Build contracts
         run: |
           forge --version
-          forge build --sizes
+          forge build
```
davidlaprade, line 29: The build step will always fail because CometFlexVoting is 8kB over the size limit

mds1: I updated the description of #47 to note that we should also bring back this size check as part of that issue

.github/workflows/ci.yml
```diff
@@ -38,32 +39,32 @@ jobs:
...
+  #         fromGitHubReleases: true
+  #         version: latest
+  #         pathInArchive: scopelint-x86_64-linux/scopelint
+  #         urlTemplate: https://github.com/ScopeLift/scopelint/releases/download/{{version}}/scopelint-x86_64-linux.tar.xz
+  #         token: ${{ secrets.GITHUB_TOKEN }}
+  #
+  #     - name: Check formatting
+  #       run: |
+  #         scopelint --version
+  #         scopelint check
```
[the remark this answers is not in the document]
mds1: We should instead replace this with a regular `forge fmt --check`

src/interfaces/IFractionalGovernor.sol
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: UNLICENSED
-pragma solidity 0.8.10;
+pragma solidity >=0.8.10;
 
 /// @dev The interface that flexible voting-compatbile governors are expected to support.
```
mds1, line 4:
```suggestion
/// @dev The interface that flexible voting-compatible governors are expected to support.
```

Also another typo `overriden` but I can't comment on that one

src/ATokenFlexVoting.sol
```diff
@@ -13,6 +13,8 @@ import {SafeCast} from "@openzeppelin/contracts/utils/math/SafeCast.sol";
 import {Checkpoints} from "@openzeppelin/contracts/utils/Checkpoints.sol";
 import {IFractionalGovernor} from "src/interfaces/IFractionalGovernor.sol";
 import {IVotingToken} from "src/interfaces/IVotingToken.sol";
+
+import {FlexVotingClient} from "src/FlexVotingClient.sol";
```
mds1, line 17: nit, feels like a weird line break location, maybe you meant this, i.e. line break after the deps end and before `src/` starts?

```suggestion

import {IFractionalGovernor} from "src/interfaces/IFractionalGovernor.sol";
import {IVotingToken} from "src/interfaces/IVotingToken.sol";
import {FlexVotingClient} from "src/FlexVotingClient.sol";
```

src/FlexVotingClient.sol
```diff
@@ -0,0 +1,169 @@
...
+
+  /// @notice Map proposalId to an address to whether they have voted on this proposal.
+  mapping(uint256 => mapping(address => bool)) private proposalVotersHasVoted;
+
+  /// @notice Map proposalId to vote totals expressed on this proposal.
+  mapping(uint256 => ProposalVote) public proposalVotes;
+
+  /// @notice The governor contract associated with this governance token. It
+  /// must be one that supports fractional voting, e.g. GovernorCountingFractional.
+  IFractionalGovernor public immutable GOVERNOR;
```
mds1, line 37: Are we sure this can be safely immutable? What if a protocol migrates to a new governor contract and transfers ownership? RR

src/FlexVotingClient.sol
```diff
@@ -0,0 +1,169 @@
...
+  }
+
+  /// @notice Data structure to store vote preferences expressed by depositors.
+  struct ProposalVote {
+    uint128 againstVotes;
+    uint128 forVotes;
+    uint128 abstainVotes;
+  }
+
+  /// @notice Map proposalId to an address to whether they have voted on this proposal.
```
mds1, line 29: we usually use `/// @notice` is for user-facing/public/external, and `/// @dev` for internal/private stuff

src/FlexVotingClient.sol
```diff
@@ -0,0 +1,169 @@
...
+
+  /// @notice Map proposalId to vote totals expressed on this proposal.
+  mapping(uint256 => ProposalVote) public proposalVotes;
+
+  /// @notice The governor contract associated with this governance token. It
+  /// must be one that supports fractional voting, e.g. GovernorCountingFractional.
+  IFractionalGovernor public immutable GOVERNOR;
+
+  /// @notice Mapping from address to stored (not rebased) balance checkpoint history.
+  mapping(address => Checkpoints.History) private balanceCheckpoints;
```
mds1, line 40: should this comment be tweaked? e.g. seems like it might be a bit odd when this is implemented for something that doesn't have rebase tokens

src/FlexVotingClient.sol
```diff
@@ -0,0 +1,169 @@
...
+
+  /// @notice History of total stored (not rebased) balances.
```
mds1, line 42: similar comments, let's check for and clarify "rebase" word usage + notice/dev usage throughout

src/FlexVotingClient.sol
```diff
@@ -0,0 +1,169 @@
...
+    GOVERNOR = IFractionalGovernor(_governor);
+  }
+
+  /// @notice Returns the _user's current balance in storage. If the balance
+  /// rebases this should return the non-rebased value, i.e. the value before
+  /// any computations are run or interest is applied.
+  function _rawBalanceOf(address _user) internal view virtual returns (uint256);
+
+  // TODO add natspec
+  function selfDelegate() public {
```
mds1, line 57: should this be `internal` because it's called at initialization/construction? not clear why it should be public

src/CometFlexVoting.sol
```diff
@@ -0,0 +1,52 @@
...
+
+import {IVotingToken} from "src/interfaces/IVotingToken.sol";
+import {IFractionalGovernor} from "src/interfaces/IFractionalGovernor.sol";
+import {FlexVotingClient} from "src/FlexVotingClient.sol";
+
+// TODO add description
+contract CometFlexVoting is Comet, FlexVotingClient {
+  using Checkpoints for Checkpoints.History;
+
+  /// @dev Constructor.
```
mds1, line 16: I think we have this in the other contract too, personally I think we can delete this line, feels like it adds noise and no value
```suggestion
```

src/CometFlexVoting.sol
```diff
@@ -0,0 +1,52 @@
...
+  /// @param _config The configuration struct for this Comet instance.
+  /// @param _governor The address of the flex-voting-compatible governance contract.
+  constructor(CometConfiguration.Configuration memory _config, address _governor)
+    Comet(_config)
+    FlexVotingClient(_governor)
+  {
+    selfDelegate();
+  }
+
+  /// @notice Returns the _user's current balance in storage.
```
mds1, line 26: var name is wrong in this comment (also added backticks since natspec comments should be written as markdown since `forge doc` will parse the markdown)
```suggestion
  /// @notice Returns the current balance in storage for the `account`.
```

src/CometFlexVoting.sol
```diff
@@ -0,0 +1,52 @@
...
+  function _rawBalanceOf(address account) internal view override returns (uint256) {
+    int104 _principal = userBasic[account].principal;
+    return _principal > 0 ? uint256(int256(_principal)) : 0;
```
mds1, line 29: why is this conditional needed? since `uint256(int256(int104(0)))` equals 0. Also this should have a safe cast to ensure the int value fits in uint. RR
```suggestion
    return toUint256(int256(userBasic[account].principal));
```

where `toUint256` is something like the solmate or OZ safecast methods

src/CometFlexVoting.sol
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  // forgefmt: disable-start
+  //===========================================================================
+  // BEGIN: Comet overrides
+  //===========================================================================
+  //
+// This function is called any time the underlying balance is changed.
+  function updateBasePrincipal(
```
mds1, line 38:
```suggestion
  //
  // This function is called any time the underlying balance is changed.
  function updateBasePrincipal(
```

src/CometFlexVoting.sol
```diff
@@ -0,0 +1,52 @@
...
+    UserBasic memory _userBasic,
+    int104 _principalNew
+  ) internal override {
+    Comet.updateBasePrincipal(_account, _userBasic, _principalNew);
+    FlexVotingClient._checkpointRawBalanceOf(_account);
+    FlexVotingClient.totalDepositCheckpoints.push(uint224(totalSupplyBase));
+  }
+  //
+//===========================================================================
+  // END: Comet overrides
```
mds1, line 49:
```suggestion
  //
  //===========================================================================
  // END: Comet overrides
```

src/CometFlexVoting.sol
```diff
@@ -0,0 +1,52 @@
...
+    selfDelegate();
+  }
+
+  /// @notice Returns the _user's current balance in storage.
+  function _rawBalanceOf(address account) internal view override returns (uint256) {
+    int104 _principal = userBasic[account].principal;
+    return _principal > 0 ? uint256(int256(_principal)) : 0;
+  }
+
+  // forgefmt: disable-start
```
mds1, line 32: I think we can prob leave forge fmt enabled here right? we're overriding the methods and calling the parent ones so "not using forge fmt to simplify the diff" seems not applicable

src/CometFlexVoting.sol
davidlaprade: Weird, forge fmt will fail without this, it was introduced here https://github.com/ScopeLift/flexible-voting/pull/46/commits/2d2908c62d03336943a61f6ce85c96e9fea160d5

src/CometFlexVoting.sol
davidlaprade: Same weird forge fmt thing

test/CometFlexVotingFork.t.sol
```diff
@@ -0,0 +1,1036 @@
...
+
+import {CometConfiguration} from "comet/CometConfiguration.sol";
+import {Comet} from "comet/Comet.sol";
+
+contract CometForkTest is Test, CometConfiguration {
+  uint256 forkId;
+
+  CometFlexVoting cToken;
+  // The Compound governor, not to be confused with the govToken's governance system:
+  address immutable COMPOUND_GOVERNOR = 0x6d903f6003cca6255D85CcA4D3B5E5146dC33925;
```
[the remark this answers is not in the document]
davidlaprade: > I also don't understand what you mean here

You do. You got it all right!

> I'm unclear what receiver will be

The receiver is just the contract that serves as the proposal target. When we create proposals on the governor, they are proposals to call a function on the receiver contract. I tried to document it [here](https://github.com/ScopeLift/flexible-voting/pull/46/files#diff-2205137817cc9796d3b572dcc15220b4b9c1df45eb62830c3f8e021f6e2cc68bR66). Happy to add more details if you think I need it

> I'd prob rename flexVotingGovernor to govTokenGovernor to make the relationship more clear

Curious what @apbendi thinks about renaming. Happy to add more comments to clarify, though

src/CometFlexVoting.sol
davidlaprade: > why is this conditional needed?

Because `_principal` can be negative, e.g. if someone has borrowed the base asset

> Also this should have a safe cast to ensure the int value fits in uint

Why do we need a safe cast? If the value fits in `int104` and it's positive then we know it fits in `uint256`

test/CometFlexVotingFork.t.sol
```diff
@@ -0,0 +1,1036 @@
...
+      18, // decimals
+      790_000_000_000_000_000, // borrowCollateralFactor
+      850_000_000_000_000_000, // liquidateCollateralFactor
+      930_000_000_000_000_000, // liquidationFactor
+      1_250_000_000_000_000_000_000_000 // supplyCap
+    );
+    Configuration memory _config = Configuration(
+      COMPOUND_GOVERNOR,
+      0xbbf3f1421D886E9b2c5D716B5192aC998af2012c, // pauseGuardian
+      address(govToken), // baseToken
```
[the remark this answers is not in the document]
davidlaprade: Lent to earn yield and borrowed, yes

test/CometFlexVotingFork.t.sol
```diff
@@ -0,0 +1,1036 @@
...
+      5_000_000_000_000, // targetReserves
+      _assetConfigs
+    );
+
+    cToken = new CometFlexVoting(_config, address(flexVotingGovernor));
+
+    cToken.initializeStorage();
+    // ========= END DEPLOY NEW COMET ========================
+
+    // TODO is there anything we need to do to make this an "official" Comet?
```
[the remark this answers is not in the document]
davidlaprade: Good idea 👍

test/CometFlexVotingFork.t.sol
```diff
@@ -0,0 +1,1036 @@
...
+    govToken.exposed_mint(_who, _govAmount);
+    vm.startPrank(_who);
+    govToken.approve(address(cToken), type(uint256).max);
+    cToken.supply(address(govToken), _govAmount);
+    vm.stopPrank();
+  }
+
+  function _createAndSubmitProposal() internal returns (uint256 proposalId) {
+    // Proposal will underflow if we're on the zero block.
+    if (block.number == 0) vm.roll(42);
```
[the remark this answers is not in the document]
davidlaprade: Good point, I'll remove this

test/CometFlexVotingFork.t.sol
```diff
@@ -0,0 +1,1036 @@
...
+    bytes memory receiverCallData = abi.encodeWithSignature("mockReceiverFunction()");
+    address[] memory targets = new address[](1);
+    uint256[] memory values = new uint256[](1);
+    bytes[] memory calldatas = new bytes[](1);
+    targets[0] = address(receiver);
+    values[0] = 0; // no ETH will be sent
+    calldatas[0] = receiverCallData;
+
+    // Submit the proposal.
+    proposalId = flexVotingGovernor.propose(targets, values, calldatas, "A great proposal");
```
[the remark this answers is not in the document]
davidlaprade: > Or is the intent of this method that the test contract is always proposing?

Correct, the test contract is always proposing. This works because the proposal threshold is [0 by default](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/96b95592c300a1b38953b276f0b6c77017f5c942/contracts/governance/Governor.sol#L199-L204), so the proposer doesn't need to have any govTokens

I'll add a comment about this. I agree it's not obvious

test/CometFlexVotingFork.t.sol
```diff
@@ -0,0 +1,1036 @@
...
+    assertEq(uint256(flexVotingGovernor.state(proposalId)), uint256(ProposalState.Pending));
+
+    // advance proposal to active state
+    vm.roll(flexVotingGovernor.proposalSnapshot(proposalId) + 1);
+    assertEq(uint256(flexVotingGovernor.state(proposalId)), uint256(ProposalState.Active));
```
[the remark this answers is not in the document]
davidlaprade: You're right, I could be better about this. Thanks

src/CometFlexVoting.sol
mds1, line 1: there's lots of compiler warnings, but most come from dependencies. the only one that doesn't is:

```
warning[3628]: Warning: This contract has a payable fallback function, but no receive ether function. Consider adding a receive ether function.
  --> lib/comet/contracts/Comet.sol:13:1:
   |
13 | contract Comet is CometMainInterface {
   | ^ (Relevant source part starts here and spans across multiple lines).
Note: The payable fallback function is defined here.
    --> lib/comet/contracts/Comet.sol:1314:5:
     |
1314 |     fallback() external payable {
     |     ^ (Relevant source part starts here and spans across multiple lines).



warning[3628]: Warning: This contract has a payable fallback function, but no receive ether function. Consider adding a receive ether function.
  --> src/CometFlexVoting.sol:13:1:
   |
13 | contract CometFlexVoting is Comet, FlexVotingClient {
   | ^ (Relevant source part starts here and spans across multiple lines).
Note: The payable fallback function is defined here.
    --> lib/comet/contracts/Comet.sol:1314:5:
     |
1314 |     fallback() external payable {
     |     ^ (Relevant source part starts here and spans across multiple lines).
```

but as you can see, I think it's just something we inherited from comet? so I don't think there's anything actionable here, just wanted to call it out

src/CometFlexVoting.sol
mds1: ah yes that all makes sense! though it seems we're not actually using the `_rawBalanceOf` methods anywhere, should they all be removed?

test/CometFlexVotingFork.t.sol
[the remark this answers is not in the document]
mds1: > You do. You got it all right!

Woohoo! 

That link doesn't seem to be taking me to the right place, but either way maybe `proposalTarget` is a better name than `receiver`? I don't feel strongly on this comment so feel free to self-resolve or wait for feedback from ben, just wanted to call out that it was a bit confusing at first read

.github/workflows/ci.yml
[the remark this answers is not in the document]
davidlaprade: Done in 951fcdf

src/FlexVotingClient.sol
davidlaprade: Hm, that's a great point. This probably shouldn't be immutable. The abstract client should also probably require inheriting contracts to define a function to update the governor as well

Since this is a somewhat separate concern, tracking https://github.com/ScopeLift/flexible-voting/issues/48

.github/workflows/ci.yml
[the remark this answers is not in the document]
mds1: lgtm but you can remove the "Install scopelint" step to speed up that CI job a bit

src/CometFlexVoting.sol
mds1: oops somehow I missed that in my code search, resolving this! 😅

src/FlexVotingClient.sol
davidlaprade: Tweaked comments and fleshed out natspec for this file in 375b73b

src/FlexVotingClient.sol
davidlaprade: Tweaked comments and fleshed out natspec for this file in 375b73b

src/FlexVotingClient.sol
davidlaprade: Agree, good catch. Done in 82c2ba0

test/CometFlexVotingFork.t.sol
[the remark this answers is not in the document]
davidlaprade: I like proposalTarget better, thanks!

src/FlexVotingClient.sol
davidlaprade: Fixed in 1ac8bc2, thanks

test/CometFlexVotingFork.t.sol
```diff
@@ -0,0 +1,1036 @@
+// SPDX-License-Identifier: Unlicensed
+pragma solidity >=0.8.10;
+
+import {Test} from "forge-std/Test.sol";
+import {Vm} from "forge-std/Vm.sol";
+import "forge-std/console2.sol";
+
+import {IVotes} from "@openzeppelin/contracts/governance/utils/IVotes.sol";
+import {ERC20} from "@openzeppelin/contracts/token/ERC20/ERC20.sol";
```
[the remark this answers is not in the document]
davidlaprade: Done in 56bf3cf, thanks

test/CometFlexVotingFork.t.sol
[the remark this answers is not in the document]
davidlaprade: Switched to `proposalTarget` in 56bf3cf. Sticking with `flexVotingGovernor` for now unless @apbendi feels strongly about it

test/CometFlexVotingFork.t.sol
```diff
@@ -0,0 +1,1036 @@
...
+  address immutable COMPOUND_GOVERNOR = 0x6d903f6003cca6255D85CcA4D3B5E5146dC33925;
+  GovToken govToken;
+  FractionalGovernor flexVotingGovernor;
+  ProposalReceiverMock receiver;
+
+  // Mainnet addresses.
+  address weth = 0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2;
+
+  // See CometMainInterface.sol.
+  error NotCollateralized();
```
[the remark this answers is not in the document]
davidlaprade: Done in 56bf3cf, thanks

test/CometFlexVotingFork.t.sol
```diff
@@ -0,0 +1,1036 @@
...
+  enum VoteType {
+    Against,
+    For,
+    Abstain
+  }
+
+  function setUp() public {
+    // Compound v3 has been deployed to mainnet.
+    // https://docs.compound.finance/#networks
+    uint256 mainnetForkBlock = 17_146_483;
```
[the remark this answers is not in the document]
davidlaprade: We do rely on a bunch of other deployed contracts (e.g. governors, chainlink price feeds, token addresses) which could conceivably change, but I think it's quite unlikely that changes to these would affect the assumptions of the tests.

Mostly I think we can just use the latest block. Added a note about this in 56bf3cf

test/CometFlexVotingFork.t.sol
```diff
@@ -0,0 +1,1036 @@
...
+    // ========= END DEPLOY NEW COMET ========================
+
+    // TODO is there anything we need to do to make this an "official" Comet?
+  }
+
+  // ------------------
+  // Helper functions
+  // ------------------
+
+  function _mintGovAndSupplyToCompound(address _who, uint256 _govAmount) internal {
```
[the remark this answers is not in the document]
davidlaprade: Fine with me, done in 56bf3cf

test/CometFlexVotingFork.t.sol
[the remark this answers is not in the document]
davidlaprade: Done in 56bf3cf, thanks

test/CometFlexVotingFork.t.sol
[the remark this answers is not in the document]
davidlaprade: Done in 56bf3cf, thanks

test/CometFlexVotingFork.t.sol
```diff
@@ -0,0 +1,1036 @@
...
+    assertEq(uint256(flexVotingGovernor.state(proposalId)), uint256(ProposalState.Pending));
+
+    // advance proposal to active state
+    vm.roll(flexVotingGovernor.proposalSnapshot(proposalId) + 1);
+    assertEq(uint256(flexVotingGovernor.state(proposalId)), uint256(ProposalState.Active));
+  }
+}
+
+contract Setup is CometForkTest {
+  function testFork_SetupCTokenDeploy() public {
```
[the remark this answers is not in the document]
davidlaprade: Done in 56bf3cf, thanks

test/CometFlexVotingFork.t.sol
```diff
@@ -0,0 +1,1036 @@
...
+    cToken.withdraw(address(govToken), 100 ether);
+    assertEq(govToken.balanceOf(_borrower), 100 ether);
+    uint256 _initBorrowBalance = cToken.borrowBalanceOf(_borrower);
+    assertEq(_initBorrowBalance, 100 ether);
+
+    // Supplier earns yield. Borrowerer owes interest.
+    vm.roll(block.number + 1);
+    vm.warp(block.timestamp + 100 days);
+    uint256 _newCTokenBalance = cToken.balanceOf(_supplier);
+    assertTrue(_newCTokenBalance > _initCTokenBalance, "Supplier has not earned yield");
```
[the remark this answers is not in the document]
davidlaprade: Done in 56bf3cf, thanks

test/CometFlexVotingFork.t.sol
```diff
@@ -0,0 +1,1036 @@
...
+    );
+  }
+
+  function test_UserCannotDoubleVoteAfterVotingAbstain() public {
+    _tesNoDoubleVoting(
+      makeAddr("test_UserCannotDoubleVoteAfterVoting address"), 0.042 ether, uint8(VoteType.Abstain)
+    );
+  }
+
+  function _tesNoDoubleVoting(address _who, uint256 _voteWeight, uint8 _supportType) private {
```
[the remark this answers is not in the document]
davidlaprade: Done in 56bf3cf, thanks

test/CometFlexVotingFork.t.sol
```diff
@@ -0,0 +1,1036 @@
...
+        1 ether, // borrowerAssets
+        uint8(VoteType.For), // supportTypeA
+        uint8(VoteType.Abstain) // supportTypeB
+      )
+    );
+  }
+
+  function _testVoteWeightIsScaledBasedOnPoolBalance(VoteWeightIsScaledVars memory _vars) private {
+    // This would be a vm.assume if we could do fuzz tests.
+    assertLt(_vars.voteWeightA + _vars.voteWeightB, type(uint128).max);
```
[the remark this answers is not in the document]
davidlaprade: Done in 56bf3cf, thanks

.github/workflows/ci.yml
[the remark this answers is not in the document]
davidlaprade: Ah, good point 3f1a7ae

src/ATokenFlexVoting.sol
```diff
@@ -13,6 +13,8 @@ import {SafeCast} from "@openzeppelin/contracts/utils/math/SafeCast.sol";
 import {Checkpoints} from "@openzeppelin/contracts/utils/Checkpoints.sol";
 import {IFractionalGovernor} from "src/interfaces/IFractionalGovernor.sol";
 import {IVotingToken} from "src/interfaces/IVotingToken.sol";
+
```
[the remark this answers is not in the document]
davidlaprade: Thanks! 3f1a7ae

test/CometFlexVotingFork.t.sol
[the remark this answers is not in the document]
davidlaprade: I looked into this a bit. [Here's the proposal](https://compound.finance/governance/proposals/152) where they began the migration from v2 to v3. You can see that at that point the v3 contracts had already been deployed, e.g. cUSDCv3 was deployed [~7mo](https://etherscan.io/tx/0xfe8e6819f1b43dedacf864d96beaf6b924b69796a28e8c4a5622f21b3cc83cc0) before this proposal landed (Aug '22 vs Mar '23).

The deployed contract is an OZ `TransparentUpgradeableProxy`, though, so the initial deployment tx doesn't tell us a ton about what (if anything) goes into making one of these things "official". And all [the proposal](https://compound.finance/governance/proposals/152) actually does is use the `Configurator` to set some configuration vars (supply caps, borrow speeds, price factors) and upgrade the proxy to a new contract. It's not pushing the cUSDCv3 contract to a whitelist or anything obvious that would encode it as being "official" or anything.

This is further confirmed if you look at the [`Configurator.deploy` function](https://github.com/compound-finance/comet/blob/d7b414d3dbe57f6e59c2c571b5a629ca7a6fec68/contracts/Configurator.sol#L329-L337), which is meant to deploy new Comet instances:

```solidity
    /**
     * @notice Deploy a new Comet implementation using the factory and Configuration for that Comet proxy
     * @dev Note: Callable by anyone
     */
    function deploy(address cometProxy) external returns (address) {
        address newComet = CometFactory(factory[cometProxy]).clone(configuratorParams[cometProxy]);
        emit CometDeployed(cometProxy, newComet);
        return newComet;
    }
```

This is `external` so anyone can call it. New Comet instances aren't whitelisted or anything. But you can see that the Configurator determines (a) what the proxy is and (b) what configuration new instances will have.

So basically it looks like the only thing that makes a given Comet proxy contract "official" is that it was deployed by the Configurator. Thus var there are only [14](https://dune.com/queries/2490944) "official" v3 Comet instances -- though to be honest I'm not sure why there are even that many since they all have to have identical configurations. To be able to deploy anything from the Configurator it has to be pre-approved by the [Governor](https://etherscan.io/address/0x6d903f6003cca6255D85CcA4D3B5E5146dC33925), i.e. Compound's timelock contract, in the form of adding entries to the Configurator's `factory` and `configuratorParams` mappings for the approved proxy address. As we can see from the Configurator's [transaction history](https://etherscan.io/address/0x316f9708bb98af7da9c68c1c3b5e79039cd336e3), this has only ever been done with one proxy address, cUSDCv3 (here is [the setFactory call](https://etherscan.io/tx/0xd28da29068f2141e947c233fb62f8787c54edf9c5dd253356002ebc64f8adfcf), here is [the setConfiguration call](https://etherscan.io/tx/0x74092c79d75e410f632a00155c23650bcecd800fad8ef1abd28448c3ade40b7b)).

So if we wanted to get really fancy we could do that: deploy a new proxy, prank Governor calls to `setFactory` and `setConfiguration` for it. But tbh I don't see what it gains for us.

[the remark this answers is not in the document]
mds1: Great write up, thanks. Leaving unresolved for @apbendi to see since this is useful. I agree that probably isn't necessary

test/CometFlexVotingFork.t.sol
```diff
@@ -0,0 +1,1036 @@
...
+    (uint256 _againstVotes, uint256 _forVotes, uint256 _abstainVotes) =
+      flexVotingGovernor.proposalVotes(_proposalId);
+
+    if (_supportTypeA == uint8(VoteType.For)) assertEq(_forVotes, _voteWeightA);
+    if (_supportTypeA == uint8(VoteType.Against)) assertEq(_againstVotes, _voteWeightA);
+    if (_supportTypeA == uint8(VoteType.Abstain)) assertEq(_abstainVotes, _voteWeightA);
+  }
+}
+
+// TODO
```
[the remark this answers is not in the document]
mds1: That seems reasonable to me, feel free to resolve this comment once we have that issue

mds1 (approved):

src/FlexVotingClient.sol
```diff
@@ -0,0 +1,209 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity >=0.8.10;
+
+import {SafeCast} from "@openzeppelin/contracts/utils/math/SafeCast.sol";
+import {Checkpoints} from "@openzeppelin/contracts/utils/Checkpoints.sol";
+import {IFractionalGovernor} from "src/interfaces/IFractionalGovernor.sol";
+import {IVotingToken} from "src/interfaces/IVotingToken.sol";
+
+/// @notice This is an abstract contract to make it easy to build clients for
```
apbendi, line 9: Awesome explanation. Absolutely love this abstraction for third party devs 😍

src/FlexVotingClient.sol
```diff
@@ -0,0 +1,209 @@
...
+
+    // Clear the stored votes so that we don't double-cast them.
+    delete proposalVotes[proposalId];
+
+    bytes memory fractionalizedVotes =
+      abi.encodePacked(_againstVotesToCast, _forVotesToCast, _abstainVotesToCast);
+    GOVERNOR.castVoteWithReasonAndParams(
+      proposalId,
+      unusedSupportParam,
+      "rolled-up vote from governance token holders", // Reason string.
```
apbendi, line 187: Random thought: we could pull this into a virtual method implemented by the inheritor to allow for each client to implement a custom reason string. If you'd prefer to pull this into an issue and we can discuss the idea later, that's fine.

src/CometFlexVoting.sol
```diff
@@ -0,0 +1,45 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity >=0.8.10;
+
+import {Comet} from "comet/Comet.sol";
+import {CometConfiguration} from "comet/CometConfiguration.sol";
+import {Checkpoints} from "@openzeppelin/contracts/utils/Checkpoints.sol";
+
+import {FlexVotingClient} from "src/FlexVotingClient.sol";
+
+// TODO add description
```
apbendi, line 10: Should we go ahead and flesh out the description as part of this PR?

src/CometFlexVoting.sol
```diff
@@ -0,0 +1,45 @@
...
+  function _rawBalanceOf(address account) internal view override returns (uint256) {
+    int104 _principal = userBasic[account].principal;
+    return _principal > 0 ? uint256(int256(_principal)) : 0;
+  }
+
+  //===========================================================================
+  // BEGIN: Comet overrides
+  //===========================================================================
+  //
+  // This function is called any time the underlying balance is changed.
```
apbendi, line 33: I love how small our footprint of changes is to the contract. That's largely because of your awesome work abstracting the client contract, and then because of this assertion about the Comet contracts. I'll do a bit of poking around to sanity check it's true. Do you feel very confident about it. @mds1 did you get a chance to look at it?

Overall I think this is great. I agree with the approach and agree with seeing if we can get a PR into Compound's repo to get the function marked as virtual.

apbendi (approved):
Looks great @davidlaprade! A few small things, but overall this is excellent work and I'm excited to have a clean client abstraction, in addition to the Comet implementation.

src/FlexVotingClient.sol
davidlaprade: Oh, I like that idea!

src/FlexVotingClient.sol
davidlaprade: Thank you! 🙂

src/CometFlexVoting.sol
davidlaprade: Yes, thanks for flagging

src/CometFlexVoting.sol
```diff
@@ -0,0 +1,45 @@
...
+contract CometFlexVoting is Comet, FlexVotingClient {
+  using Checkpoints for Checkpoints.History;
+
+  /// @param _config The configuration struct for this Comet instance.
+  /// @param _governor The address of the flex-voting-compatible governance contract.
+  constructor(CometConfiguration.Configuration memory _config, address _governor)
+    Comet(_config)
+    FlexVotingClient(_governor)
+  {
+    _selfDelegate();
```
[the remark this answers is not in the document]
davidlaprade: Hm, good question.

We aren't currently doing it in the constructor because if the client/pool is a proxy then it has to self delegate some other way (e.g. within an initializer). But I don't necessarily see a drawback to self delegating in both places. If a logic contract has self-delegated (because self-delegation happens in its constructor) who cares -- it won't even hold any tokens so it shouldn't matter

cc @mds1 in case there's something I'm missing here

src/CometFlexVoting.sol
davidlaprade: Thank you!

> Do you feel very confident about it

Yeah, I'm fairly certain this function is the only way the principal balance is ever changed. The balance is stored as [`userBasic[account].principal`](https://github.com/compound-finance/comet/blob/3780c06b4eaa80a8c78e0ff770a7e8a1518db75e/contracts/CometStorage.sol#L68-L69). If you search through Comet.sol and CometExt.sol for `principal =` or `userBasic[.*=` you find that the only place where the principal can change is [here](https://github.com/compound-finance/comet/blob/3780c06b4eaa80a8c78e0ff770a7e8a1518db75e/contracts/Comet.sol#L759), in `updateBasePrincipal`.

I'll open my PR with them and see if they disagree.

davidlaprade: https://github.com/compound-finance/comet/pull/749

src/CometFlexVoting.sol
davidlaprade: Done in c3b2fe3

src/CometFlexVoting.sol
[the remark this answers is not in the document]
mds1: Just for reference the method looks like this:

```solidity
/// @dev Delegates the present contract's voting rights with `GOVERNOR` to itself.
function _selfDelegate() internal {
  IVotingToken(GOVERNOR.token()).delegate(address(this));
}
```

And we set the `GOVERNOR` in the constructor of `FlexVotingClient`:

```solidity
/// @param _governor The address of the flex-voting-compatible governance contract.
constructor(address _governor) {
  GOVERNOR = IFractionalGovernor(_governor);
}
```

 So IF we called `_selfDelegate` in the constructor of `FlexVotingClient` AND someone wanted to deploy a logic (implementation) contract for a proxy that used `FlexVotingClient`, THEN they'd have to set a real governor in their logic contract, otherwise the constructor would revert because the call would fail. Personally I like using dummy/null values to be set in logic contracts, as it makes their purpose more clear.

It seems we could either:
1. Leave things as-is, this gives implementers the most optionality
2. Update `_selfDelegate` to be a no-op if the governor address is the zero address (limits what the "dummy value" for the governor can be in the logic contract)
3. Update `_selfDelegate` to have a try/catch so we can call it in the constructor and let people use any dummy values in their logic contracts

Option 1 seems the simplest to me.

But to this point, aren't we currently inconsistent with regard to proxies? You can only set `GOVERNOR` in the constructor, so we don't support proxies nicely anyway, i.e. you can't have a proxy factory that spits out flex voting clients with different governors. So maybe the `FlexVotingClient` needs to replace the constructor with an `initialize` method that both sets the governor and then self-delegates.

src/CometFlexVoting.sol
mds1: I have not looked closely at coment, but I can spend some time reviewing to confirm the assumption if you want

src/FlexVotingClient.sol
davidlaprade: Done in 9228936, thanks for the suggestion

src/CometFlexVoting.sol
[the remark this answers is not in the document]
davidlaprade: Great point about proxies -- added this motivation to our existing ticket to make `GOVERNOR` mutable, https://github.com/ScopeLift/flexible-voting/issues/48

I agree that we should just leave as-is for now

src/CometFlexVoting.sol
davidlaprade: Up to you guys. We can also just wait for the Compound eng people to get back to us

.gitmodules
```diff
@@ -11,3 +11,6 @@
 	path = lib/openzeppelin-contracts
 	url = https://github.com/openzeppelin/openzeppelin-contracts
 	branch = v4.8.0
+[submodule "lib/comet"]
+	path = lib/comet
+	url = https://github.com/davidlaprade/comet
```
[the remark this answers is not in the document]
davidlaprade: PR for upstream change is here: https://github.com/compound-finance/comet/pull/749

test/CometFlexVotingFork.t.sol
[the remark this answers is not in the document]
davidlaprade: Issue here: https://github.com/ScopeLift/flexible-voting/issues/49

Comments removed in c51734f

src/CometFlexVoting.sol
```diff
@@ -0,0 +1,77 @@
...
+
+/// @notice This is an extension of Compound V3's Comet contract which makes it
+/// possible for Comet token holders to still vote on governance proposals. This
+/// way, holders of governance tokens do not have to choose between earning
+/// yield on Compound and voting. They can do both.
+///
+/// This extension has the following requirements:
+///   (a) the base token be a governance token
+///   (b) the base token's governor contract support flexible voting (see
+///       GovernorCountingFractional)
```
mds1, line 18: grammar/natspec nits
```suggestion
///   (a) The base token must be a governance token.
///   (b) The base token's governor contract must support flexible voting (see
///       `GovernorCountingFractional`).
```

src/CometFlexVoting.sol
apbendi: I took a look and I agree with your assessment @davidlaprade. Combined with the fact that we have the PR with Compound I think we can close this.

src/CometFlexVoting.sol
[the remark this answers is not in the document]
apbendi: Hmm yeah all good points about the proxies. On the one hand I think that deploying Governors via Proxy isn't currently something I see happening, and I'm not sure it makes a lot of sense. That said, I don't think we should foreclose on working well with them. So we should consider it more carefully. Let's pull out an issue to consider Proxy support and link to this convo. Otherwise can be closed.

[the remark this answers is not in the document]
mds1: > On the one hand I think that deploying Governors via Proxy isn't currently something I see happening

Just to clarify, this is about deploying the _voting client_, not the governor, and clients are often proxies (e.g. AToken contracts)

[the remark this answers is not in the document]
apbendi: Yep good point. I was mixed up when I wrote that, thanks. We should definitely support Proxies for clients.

PR #51, opened 2023-05-30
Prevent partial vote replays

### LINKED ISSUE #50: Partial voting signature replay issue with castVoteWithReasonAndParamsBySig + GovernorCountingFractional
There is no nonce by default for `castVoteWithReasonAndParamsBySig` and `_castVote` depends on voting only once. This is fine for the normal flow, but partial votes by sig can be replayed in `GovernorCountingFractional` as long as they are still under the total voting weight of that user. Adding a nonce to `bytes memory params` and enforcing it would fix this issue.

davidlaprade:
Fixes #50

### DIFF
--- a/src/GovernorCountingFractional.sol
+++ b/src/GovernorCountingFractional.sol
@@ -3,6 +3,7 @@
 
 pragma solidity ^0.8.0;
 
+import {ECDSA} from "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
 import {Governor} from "@openzeppelin/contracts/governance/Governor.sol";
 import {GovernorCompatibilityBravo} from "@openzeppelin/contracts/governance/compatibility/GovernorCompatibilityBravo.sol";
 import {SafeCast} from "@openzeppelin/contracts/utils/math/SafeCast.sol";
@@ -19,6 +20,11 @@ import {SafeCast} from "@openzeppelin/contracts/utils/math/SafeCast.sol";
  * knowledge proofs.
  */
 abstract contract GovernorCountingFractional is Governor {
+
+    // We use this instead of EXTENDED_BALLOT_TYPEHASH to prevent the recasting
+    // of signed partial votes.
+    bytes32 public constant FRACTIONAL_BALLOT_TYPEHASH =
+        keccak256("ExtendedBallot(uint256 proposalId,uint8 support,string reason,bytes params,uint256 nonce)");
 
     struct ProposalVote {
         uint128 againstVotes;
@@ -37,6 +43,8 @@ abstract contract GovernorCountingFractional is Governor {
      * would tell you the number of votes that 0xBEEF has cast on proposal 42.
      */
     mapping(uint256 => mapping(address => uint128)) private _proposalVotersWeightCast;
+
+    mapping(address => uint256) public nonce;
 
     /**
      * @dev See {IGovernor-COUNTING_MODE}.
@@ -240,5 +248,56 @@ abstract contract GovernorCountingFractional is Governor {
             forVotes := and(_VOTEMASK, mload(add(voteData, 0x20)))
             abstainVotes := shr(128, mload(add(voteData, 0x40)))
         }
+    }
+
+
+    /**
+     * @dev See {IGovernor-castVoteWithReasonAndParamsBySig}.
+     */
+    function castVoteWithReasonAndParamsBySig(
+        uint256 proposalId,
+        uint8 support,
+        string calldata reason,
+        bytes memory params,
+        uint8 v,
+        bytes32 r,
+        bytes32 s
+    ) public virtual override returns (uint256) {
+      // This function is not safe.
+      revert("Missing voter param");
+    }
+
+    function castVoteWithReasonAndParamsBySig(
+        uint256 proposalId,
+        uint8 support,
+        string calldata reason,
+        bytes memory params,
+        address voter,
+        uint8 v,
+        bytes32 r,
+        bytes32 s
+    ) public virtual returns (uint256) {
+        address _recoveredVoter = ECDSA.recover(
+            _hashTypedDataV4(
+                keccak256(
+                    abi.encode(
+                        FRACTIONAL_BALLOT_TYPEHASH,
+                        proposalId,
+                        support,
+                        keccak256(bytes(reason)),
+                        keccak256(params),
+                        nonce[voter]
+                    )
+                )
+            ),
+            v,
+            r,
+            s
+        );
+
+        if (_recoveredVoter != voter) revert("Message signature mismatch");
+
+        nonce[voter]++;
+        return _castVote(proposalId, voter, support, reason, params);
     }
 }
--- a/src/GovernorCountingFractional.sol
+++ b/src/GovernorCountingFractional.sol
@@ -21,11 +21,6 @@ import {SafeCast} from "@openzeppelin/contracts/utils/math/SafeCast.sol";
  */
 abstract contract GovernorCountingFractional is Governor {
 
-    // We use this instead of EXTENDED_BALLOT_TYPEHASH to prevent the recasting
-    // of signed partial votes.
-    bytes32 public constant FRACTIONAL_BALLOT_TYPEHASH =
-        keccak256("FractionalBallot(uint256 proposalId,uint8 support,string reason,bytes params,uint256 nonce)");
-
     struct ProposalVote {
         uint128 againstVotes;
         uint128 forVotes;
@@ -44,7 +39,8 @@ abstract contract GovernorCountingFractional is Governor {
      */
     mapping(uint256 => mapping(address => uint128)) private _proposalVotersWeightCast;
 
-    mapping(address => uint256) public nonce;
+    // TODO add natspec
+    mapping(address => uint128) public nonces;
 
     /**
      * @dev See {IGovernor-COUNTING_MODE}.
@@ -253,40 +249,56 @@ abstract contract GovernorCountingFractional is Governor {
 
     /**
      * @dev See {IGovernor-castVoteWithReasonAndParamsBySig}.
+     * TODO add new docs
+     * the final 128bits of the params should contain the nonce.
      */
-    function castVoteWithReasonAndParamsBySig(
-        uint256,
-        uint8,
-        string calldata,
-        bytes memory,
-        uint8,
-        bytes32,
-        bytes32
-    ) public virtual override returns (uint256) {
-      // This function is not safe.
-      revert("Missing voter param");
-    }
-
     function castVoteWithReasonAndParamsBySig(
         uint256 proposalId,
         uint8 support,
         string calldata reason,
         bytes memory params,
-        address voter,
         uint8 v,
         bytes32 r,
         bytes32 s
-    ) public virtual returns (uint256) {
-        address _recoveredVoter = ECDSA.recover(
+    ) public virtual override returns (uint256) {
+        // Signature-based voting requires `params` be two full words in length:
+        //   16 bytes for againstVotes.
+        //   16 bytes for forVotes.
+        //   16 bytes for abstainVotes.
+        //   16 bytes for the signature nonce.
+        require(
+          params.length == 64,
+          "GovernorCountingFractional: invalid params for signature-based vote"
+        );
+
+        // Get the nonce out of the params. It is the last half-word.
+        uint128 nonce;
+        assembly {
+          nonce := and(
+            // Perform bitwise AND operation on the data in the second word of
+            // `params` with a mask of 128 zeros followed by 128 ones, i.e. take
+            // the last 128 bits of `params`.
+            _VOTEMASK, // TODO rename this to something more general.
+            // Load the data from memory at the returned address.
+            mload(
+              // Skip the first 64 bytes (i.e. 0x40):
+              //   32 bytes for the length of the bytes array.
+              //   32 bytes for the first word in the params
+              // Return the memory address for the last word in params.
+              add(params, 0x40)
+            )
+          )
+        }
+
+        address voter = ECDSA.recover(
             _hashTypedDataV4(
                 keccak256(
                     abi.encode(
-                        FRACTIONAL_BALLOT_TYPEHASH,
+                        EXTENDED_BALLOT_TYPEHASH,
                         proposalId,
                         support,
                         keccak256(bytes(reason)),
-                        keccak256(params),
-                        nonce[voter]
+                        keccak256(params)
                     )
                 )
             ),
@@ -295,9 +307,17 @@ abstract contract GovernorCountingFractional is Governor {
             s
         );
 
-        if (_recoveredVoter != voter) revert("Message signature mismatch");
+        require(
+          nonces[voter] == nonce,
+          "GovernorCountingFractional: signature has already been used"
+        );
 
-        nonce[voter]++;
+        nonces[voter]++;
+        // Trim params to only keep the first 48 bytes, which are the voting params.
+        assembly {
+          mstore(params, 48)
+        }
         return _castVote(proposalId, voter, support, reason, params);
     }
+
 }
--- a/src/GovernorCountingFractional.sol
+++ b/src/GovernorCountingFractional.sol
@@ -317,11 +317,11 @@ abstract contract GovernorCountingFractional is Governor {
               _MASK_HALF_WORD_RIGHT,
               // Load the data from memory at the returned address.
               mload(
-                // Skip the first 64 bytes (i.e. 0x40):
-                //   32 bytes for the length of the bytes array.
+                // Skip the first 64 bytes:
+                //   32 bytes encoding the length of the bytes array.
                 //   32 bytes for the first word in the params
                 // Return the memory address for the last word in params.
-                add(params, 0x40)
+                add(params, 64)
               )
             )
           }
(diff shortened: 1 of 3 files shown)


src/GovernorCountingFractional.sol
```diff
@@ -241,4 +245,79 @@ abstract contract GovernorCountingFractional is Governor {
...
         }
     }
+
+
+    /**
+     * @dev See {IGovernor-castVoteWithReasonAndParamsBySig}.
+     * TODO add new docs
+     * the final 128bits of the params should contain the nonce.
+     */
+    function castVoteWithReasonAndParamsBySig(
```
jonwalch, line 255: Completely unsolicited feedback, I prefer this impl because it sticks to OZ Gov standard 👍

jonwalch: Thanks again for the quick work on this issue!

davidlaprade: Of course, thanks for letting us know about it. Appreciate the feedback!

test/FractionalGovernor.sol
```diff
@@ -7,6 +7,20 @@ import "@openzeppelin/contracts/governance/extensions/GovernorVotes.sol";
...
+  function castVoteWithReasonAndParamsBySig(
+    uint256 proposalId,
+    uint8 support,
+    string calldata reason,
+    bytes memory params,
+    uint8 v,
+    bytes32 r,
+    bytes32 s
+  ) public virtual override(GovernorCountingFractional, Governor) returns (uint256) {
+    return GovernorCountingFractional.castVoteWithReasonAndParamsBySig(
```
mds1, line 19: Why is this method/override required? I'd have thought that by default, since we inherit from `GovernorCountingFractional` which inherits from `Governor`, calling `castVoteWithReasonAndParamsBySig` would use the `GovernorCountingFractional` version

test/GovernorCountingFractional.t.sol
```diff
@@ -377,6 +377,7 @@ contract GovernorCountingFractionalTest is Test {
     uint256 _privateKey;
     (_voter.addr, _privateKey) = makeAddrAndKey("voter");
     vm.assume(_voter.addr != address(this));
+    uint128 _initNonce = governor.nonces(_voter.addr);
```
mds1, line 380: This results in always fuzzing with a nonce of 0, we should consider adding a `nonce` input arg then using the StdStorage lib to overwrite the expected initial nonce

src/GovernorCountingFractional.sol
```diff
@@ -38,6 +39,14 @@ abstract contract GovernorCountingFractional is Governor {
      */
     mapping(uint256 => mapping(address => uint128)) private _proposalVotersWeightCast;
 
+    /**
+     * @dev Mapping from voter address to signature-based vote nonce. The
+     * voter's nonce increments each time a signature-based vote is cast with
+     * fractional voting params and must be included in the `params` as the last
+     * 32 bytes when signing for a fractional vote.
+     */
+    mapping(address => uint128) public nonces;
```
apbendi, line 48: RR I think we should consider a more descriptive name here. I also wonder if this should be singular, instead of plural, because of how it will look at the callsite for an integrator. So for example, a name like `nextFractionalVoteNonce` or something similar.

To demonstrate what it would read like for an integrator:

```solidity
uint128 _nonce = governor.nonces(my_address);
```

vs.

```solidity
uint128 _nonce = governor.nextFractionalVoteNonce(my_address);
```

What do you think?

src/GovernorCountingFractional.sol
```diff
@@ -219,7 +228,7 @@ abstract contract GovernorCountingFractional is Governor {
         _proposalVotes[proposalId] = _proposalVote;
     }
 
-    uint256 constant internal _VOTEMASK = 0xffffffffffffffffffffffffffffffff; // 128 bits of 0's, 128 bits of 1's
+    uint256 constant internal _HALF_WORD_RIGHT = 0xffffffffffffffffffffffffffffffff; // 128 bits of 0's, 128 bits of 1's
```
apbendi, line 231: nit: not actually a huge fan of removing "mask" from the name here, especially since we use it within assembly, it could be hard for someone to immediately recognize what we're doing with it. Whereas if you see a name with "mask" it's pretty easy to guess we're using it to mask off some bits.

src/GovernorCountingFractional.sol
```diff
@@ -237,8 +246,101 @@ abstract contract GovernorCountingFractional is Governor {
...
+        //   16 bytes for forVotes.
+        //   16 bytes for abstainVotes.
+        //   16 bytes for the signature nonce.
+        // Signature-based nominal voting requires `params` be 0 bytes.
+        require(
+          params.length == 64 || params.length == 0,
+          "GovernorCountingFractional: invalid params for signature-based vote"
+        );
+
+        address voter = ECDSA.recover(
```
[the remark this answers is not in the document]
apbendi: I don't think that would work, because we'd end up sending the wrong number of bytes to `_castVote` in that scenario.

test/FractionalGovernor.sol
apbendi: We also inherit from `GovernorVotes` which has its own implementation because it inherits it from `Governor`, so the compiler has no way of knowing which implementation we intend here.

apbendi (changes requested):
Overall looks great @davidlaprade, thanks for tackling this so quickly. I have one borderline nit on the naming for the nonces mapping. Sorry to be annoying about it, but I would like to make sure we discuss/re-review it because it will be important for integrators.

src/GovernorCountingFractional.sol
mds1: oh good point I agree it should be singular, but I'm indifferent about naming the var `nonce` vs. `nextFractionalVoteNonce` or `fractionalVoteNonce`

I think "next" is a bit ambiguous since this is really the current nonce you are supposed to use, whereas "next" can be interpreted as "use this next" or "what the next one will be after you use the current"

src/GovernorCountingFractional.sol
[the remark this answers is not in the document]
mds1: true but that would be easy to fix, I think the bigger issue is that `Governor. castVoteWithReasonAndParamsBySig` doesn't return the recovered voter address so how do we verify the nonce without re-recovering the sig?

test/FractionalGovernor.sol
mds1: hm, feels like a weird architecture that we have two base contracts that each inherit from governor, why is that?
- `GovernorCountingFractional is Governor`
- `GovernorVotes is Governor`
- `FractionalGovernor is GovernorVotes, GovernorCountingFractional`

src/GovernorCountingFractional.sol
apbendi: Fair point on "next". I'm fine with `fractionalVoteNonce` if @davidlaprade agrees.

test/FractionalGovernor.sol
apbendi: That's how OZ is architected. You select several extensions that implement their portion of the logic, but all inherit from `Governor`. I don't remember it off the top of my head, but iirc there is a name for this as a OOP pattern.

mds1: got it, that does make sense for an arch where users need to be choose multiple extensions, will mark this resolved

src/GovernorCountingFractional.sol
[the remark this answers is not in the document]
davidlaprade: Yeah, I think we'd have to override `Governor.castVoteWithReasonAndParamsBySig` in order to get the recovered voter, in which case we lose the benefit of calling it

test/FractionalGovernor.sol
jonwalch: > That's how OZ is architected. You select several extensions that implement their portion of the logic, but all inherit from `Governor`. I don't remember it off the top of my head, but iirc there is a name for this as a OOP pattern.

We found this pattern to be pretty painful in Frax Governance where we use 6 extensions 😄

src/GovernorCountingFractional.sol
apbendi: sgtm

test/FractionalGovernor.sol
davidlaprade: Good question. Tbh I'm not sure why they did it that way. Maybe someone from OZ can weigh in here

But if you look at their mocks you can see that this kind of duplication is intended, e.g.:
* [GovernorVoteMock](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/30256fa8384cd8e95de7413454bb74c913ed9f3f/contracts/mocks/governance/GovernorVoteMock.sol#L8)
* [GovernorWithParamsMock](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/30256fa8384cd8e95de7413454bb74c913ed9f3f/contracts/mocks/governance/GovernorWithParamsMock.sol#L8)

We're just following their pattern

src/GovernorCountingFractional.sol
```diff
@@ -38,6 +39,14 @@ abstract contract GovernorCountingFractional is Governor {
      */
     mapping(uint256 => mapping(address => uint128)) private _proposalVotersWeightCast;
 
+    /**
+     * @dev Mapping from voter address to signature-based vote nonce. The
+     * voter's nonce increments each time a signature-based vote is cast with
+     * fractional voting params and must be included in the `params` as the last
+     * 32 bytes when signing for a fractional vote.
```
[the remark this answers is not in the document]
davidlaprade: Ah, good catch, thanks!

src/GovernorCountingFractional.sol
```diff
@@ -237,8 +246,101 @@ abstract contract GovernorCountingFractional is Governor {
...
+                //   32 bytes for the length of the bytes array.
+                //   32 bytes for the first word in the params
+                // Return the memory address for the last word in params.
+                add(params, 0x40)
+              )
+            )
+          }
+
+          require(
+            nonces[voter] == nonce,
```
jonwalch, line 331: Style comment, you could do:

`nonces[voter]++ == nonce`

But ultimately up to how you want to style your codebase.

test/GovernorCountingFractional.t.sol
```diff
@@ -463,6 +474,157 @@ contract GovernorCountingFractionalTest is Test {
...
+    uint8 v;
+    bytes32 r;
+    bytes32 s;
+    uint256 actualAgainstVotes;
+    uint256 actualForVotes;
+    uint256 actualAbstainVotes;
+    uint256 remainingWeight;
+  }
+
+  function test_RevertIf_CastVoteWithReasonAndParamsBySig_SignatureIsReplayed(
```
jonwalch, line 497: great test

src/GovernorCountingFractional.sol
```diff
@@ -237,8 +246,101 @@ abstract contract GovernorCountingFractional is Governor {
...
+            "GovernorCountingFractional: signature has already been used"
+          );
+
+          nonces[voter]++;
+
+          // Trim params to only keep the first 48 bytes, which are the voting params.
+          assembly {
+            mstore(params, 48)
```
jonwalch, line 339: I'd consider making a new memory variable to make it a bit more clear that you're passing something mutated to _castVote, however I'd imagine this impl is slightly more gas efficient.

src/GovernorCountingFractional.sol
jonwalch: I enjoy named mappings, but I know OZ doesn't use them.

Something like:
`mapping(address signer => uint128 nonce) public nonces;`

src/GovernorCountingFractional.sol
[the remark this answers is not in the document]
davidlaprade: a61d7b9

src/GovernorCountingFractional.sol
```diff
@@ -237,8 +246,101 @@ abstract contract GovernorCountingFractional is Governor {
...
         assembly {
             againstVotes := shr(128, mload(add(voteData, 0x20)))
-            forVotes := and(_VOTEMASK, mload(add(voteData, 0x20)))
+            forVotes := and(_HALF_WORD_RIGHT, mload(add(voteData, 0x20)))
             abstainVotes := shr(128, mload(add(voteData, 0x40)))
         }
     }
+
+
+    /**
```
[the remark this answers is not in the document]
davidlaprade: a61d7b9

src/GovernorCountingFractional.sol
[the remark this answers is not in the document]
davidlaprade: a61d7b9

src/GovernorCountingFractional.sol
```diff
@@ -237,8 +246,101 @@ abstract contract GovernorCountingFractional is Governor {
...
+     * @notice Cast a vote with a reason and additional encoded parameters using
+     * the user's cryptographic signature.
+     *
+     * Emits a {VoteCast} or {VoteCastWithParams} event depending on the length
+     * of params.
+     *
+     * @dev If casting a fractional vote via `params`, the voter's current nonce
+     * must be appended to the `params` and included in the signature. I.e., the
+     * params used when constructing the signature would be:
```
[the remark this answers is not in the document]
davidlaprade: a61d7b9

src/GovernorCountingFractional.sol
davidlaprade: Went with `fractionalVoteNonce`, good suggestion! Done in a61d7b9

src/GovernorCountingFractional.sol
davidlaprade: Went with that in a61d7b9

src/GovernorCountingFractional.sol
```diff
@@ -237,8 +246,101 @@ abstract contract GovernorCountingFractional is Governor {
...
+
+          nonces[voter]++;
+
+          // Trim params to only keep the first 48 bytes, which are the voting params.
+          assembly {
+            mstore(params, 48)
+          }
+        }
+
+        return _castVote(proposalId, voter, support, reason, params);
```
jonwalch, line 342: I like named function parameters, even though OZ doesn't use them:

`return _castVote({proposalId: proposalId, account: voter, support: support, reason: reason, params: params});`

test/GovernorCountingFractional.t.sol
```diff
@@ -403,6 +404,9 @@ contract GovernorCountingFractionalTest is Test {
       _proposalId, _voter.support, "I have my reasons", new bytes(0), _v, _r, _s
     );
 
+    // Nonce should not have updated.
+    assertEq(_initNonce, governor.nonces(_voter.addr));
```
jonwalch: `assertEq(_initNonce, governor.nonces(_voter.addr), "Nonce should not have updated");` is supported and the string shows up in the command line when looking at test failures. Pretty useful.

src/GovernorCountingFractional.sol
apbendi: lol ngl I didn't even know these were a thing @jonwalch!!! 🤯 

Anyway I like the change @davidlaprade and probably best to stick with the OZ convention.

mds1: Named mappings we’re introduced in 0.8.18 so we’d have to change the pragma anyway, which we should avoid (even though I also prefer named mappings)

jonwalch: Maybe too bleeding edge, but they bumped pragmas on master 4 days ago.

https://github.com/OpenZeppelin/openzeppelin-contracts/commit/7bb5592ad5f7a1f28efbc358940f251ad0841366

src/GovernorCountingFractional.sol
apbendi: Hmm good point, maybe amending the comment to say something like "trim params *in place* to keep only..." would be a good compromise.

src/GovernorCountingFractional.sol
apbendi: Will look forward to using these in the future but I think too fresh for right now. Thanks @jonwalch. Resolving this issue with @davidlaprade's change!

src/GovernorCountingFractional.sol
apbendi: Probably another one where we'll stick with OZ convention since we're extending it. I also like the named params.

src/GovernorCountingFractional.sol
davidlaprade: Same. That's elegant but I prefer the explicitness

test/GovernorCountingFractional.t.sol
davidlaprade: thanks :)

src/GovernorCountingFractional.sol
davidlaprade: Yup, going to stick with OZ's convention, but I also prefer using named parameters

test/GovernorCountingFractional.t.sol
davidlaprade: Good idea! Added in 4067f4b

test/GovernorCountingFractional.t.sol
davidlaprade: Yup, good idea. I decided to just expose a setter function on the mock governor to set the nonce. Done in 4067f4b

apbendi (approved):
Great job on this @davidlaprade. Thanks for the quick turn around.

PR #66, opened 2024-01-10
Upgrade to OZ v5

alexkeating:
#### Description

- Some errors changed 
- The check-pointing contract changed
- The `castVoteWithReasonAndParamsBySig` has a nonce now so we can remove our method override.
- The `GovernorCompatibilityBravo` contract was removed so we need to add our own VoteType struct


test/FractionalGovernor.sol
```diff
@@ -1,8 +1,10 @@
 // SPDX-License-Identifier: MIT
 pragma solidity >=0.8.10;
 
-import "../src/GovernorCountingFractional.sol";
-import "@openzeppelin/contracts/governance/extensions/GovernorVotes.sol";
+import {GovernorCountingFractional} from "../src/GovernorCountingFractional.sol";
+import {GovernorVotes} from "@openzeppelin/contracts/governance/extensions/GovernorVotes.sol";
+import {IVotes} from "@openzeppelin/contracts/governance/utils/IVotes.sol";
+import {Governor} from "@openzeppelin/contracts/governance//Governor.sol";
```
wildmolasses, line 7:
```suggestion
import {Governor} from "@openzeppelin/contracts/governance/Governor.sol";
```

src/GovernorCountingFractional.sol
```diff
@@ -250,96 +258,4 @@ abstract contract GovernorCountingFractional is Governor {
...
-     *
-     * @dev If casting a fractional vote via `params`, the voter's current nonce
-     * must be appended to the `params` as the last 16 bytes and included in the
-     * signature. I.e., the params used when constructing the signature would be:
-     *
-     *   abi.encodePacked(againstVotes, forVotes, abstainVotes, nonce)
-     *
-     * See {fractionalVoteNonce} and {_castVote} for more information.
-     */
-    function castVoteWithReasonAndParamsBySig(
```
alexkeating, line 268: This was added because the v4 version did not handle replay attacks. I think the v5 uses a nonce now so we don't need to override it.

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/920225a1c7e6e749ba23b10ed076b04f454290e8/contracts/governance/Governor.sol#L577

src/FlexVotingClient.sol
```diff
@@ -92,7 +92,7 @@ abstract contract FlexVotingClient {
   /// token that `_user` has claim to in this system. It may or may not be
   /// equivalent to the withdrawable balance of `GOVERNOR`s token for `user`,
   /// e.g. if the internal representation of balance has been scaled down.
-  function _rawBalanceOf(address _user) internal view virtual returns (uint256);
```
alexkeating, line 95: Changed because there is no `Trace256`. Should we write one for OZ?

src/FlexVotingClient.sol
```diff
@@ -194,19 +194,21 @@ abstract contract FlexVotingClient {
...
+    balanceCheckpoints[_user].push(SafeCast.toUint32(block.number), _rawBalanceOf(_user));
   }
 
   /// @notice Returns the `_user`'s raw balance at `_blockNumber`.
   /// @param _user The account that's historical raw balance will be looked up.
   /// @param _blockNumber The block at which to lookup the _user's raw balance.
   function getPastRawBalance(address _user, uint256 _blockNumber) public view returns (uint256) {
-    return balanceCheckpoints[_user].getAtProbablyRecentBlock(_blockNumber);
+    uint32 key = SafeCast.toUint32(_blockNumber);
+    return balanceCheckpoints[_user].upperLookup(key);
```
alexkeating, line 205: https://docs.openzeppelin.com/contracts/5.x/api/utils#Checkpoints-upperLookup-struct-Checkpoints-Trace224-uint32-

test/FractionalPool.t.sol
```diff
@@ -115,7 +116,7 @@ contract Deployment is FractionalPoolTest {
 
 contract Deposit is FractionalPoolTest {
   function test_UserCanDepositGovTokens(address _holder, uint256 _amount) public {
-    _amount = bound(_amount, 0, type(uint224).max);
+    _amount = bound(_amount, 0, type(uint208).max);
```
alexkeating, line 119: Why this was changed. https://github.com/OpenZeppelin/openzeppelin-contracts/blob/920225a1c7e6e749ba23b10ed076b04f454290e8/contracts/token/ERC20/extensions/ERC20Votes.sol#L40

.github/workflows/ci.yml
```diff
@@ -106,7 +106,7 @@ jobs:
         uses: zgosalvez/github-actions-report-lcov@v2
         with:
           coverage-files: ./lcov.info
-          minimum-coverage: 70 # Set coverage threshold.
```
alexkeating, line 109: Should we add more tests

test/GovernorCountingFractional.t.sol
```diff
@@ -332,7 +332,15 @@ contract GovernorCountingFractionalTest is Test {
       assertEq(uint8(status), uint8(IGovernor.ProposalState.Defeated));
 
       Proposal memory _rawProposalInfo = _getSimpleProposal();
-      vm.expectRevert(bytes("Governor: proposal not successful"));
+      vm.expectRevert(
+        abi.encodeWithSelector(
+          IGovernor.GovernorUnexpectedProposalState.selector,
+          _proposalId,
+          status,
+          bytes32(1 << uint8(IGovernor.ProposalState.Succeeded))
```
alexkeating, line 340: From this line https://github.com/OpenZeppelin/openzeppelin-contracts/blob/920225a1c7e6e749ba23b10ed076b04f454290e8/contracts/governance/Governor.sol#L403

src/FlexVotingClient.sol
```diff
@@ -48,7 +48,7 @@ import {IVotingToken} from "./interfaces/IVotingToken.sol";
 /// of the rest.
 abstract contract FlexVotingClient {
   using SafeCast for uint256;
-  using Checkpoints for Checkpoints.History;
+  using Checkpoints for Checkpoints.Trace224;
```
alexkeating, line 51: Should this be a 208?

test/FractionalPool.t.sol
Amxx: Because `Votes.sol` uses `Checkpoints.Trace208` for the checkpoints. This is a consequence of the adoption of `uint48` for all clocks/timestamps, following ERC-5805

src/FlexVotingClient.sol
Amxx: Traces are a list of checkpoint. Each checkpoint contains:
- a key
- a value
such that sizeof(key) + sizeof(value) = 256

- `Trace224` is `uint224` for value, and `uint32` for key
- `Trace208` is `uint208` for value, and `uint48` for key

`Trace256` would not have any room for the key ...

src/FlexVotingClient.sol
Amxx: If you want to allign with ERC-5805, then yes

src/FlexVotingClient.sol
```diff
@@ -194,19 +194,21 @@ abstract contract FlexVotingClient {
...
-    return balanceCheckpoints[_user].getAtProbablyRecentBlock(_blockNumber);
+    uint32 key = SafeCast.toUint32(_blockNumber);
+    return balanceCheckpoints[_user].upperLookup(key);
   }
 
   /// @notice Returns the sum total of raw balances of all users at `_blockNumber`.
   /// @param _blockNumber The block at which to lookup the total balance.
   function getPastTotalBalance(uint256 _blockNumber) public view returns (uint256) {
-    return totalBalanceCheckpoints.getAtProbablyRecentBlock(_blockNumber);
+    uint32 key = SafeCast.toUint32(_blockNumber);
```
davidlaprade, line 211: Could we just make the argument a `uint32`?

davidlaprade: Nevermind.

We could change the argument type, but given that the caller is just passing along `GOVERNOR.proposalSnapshot`, which is a uint256, we'd have to end up casting it in the callers instead. So there'd be no benefit

src/FlexVotingClient.sol
davidlaprade: Switched to Trace208 in https://github.com/ScopeLift/flexible-voting/pull/66/commits/a01bf77a1f10d12b57f3a29ee07bd600ddbde973

.github/workflows/ci.yml
```diff
@@ -106,7 +106,7 @@ jobs:
...
+          minimum-coverage: 67 # Set coverage threshold.
```
apbendi, line 109: We can bump this up to 92 or so now, I think!

test/FlexVotingClient.t.sol
```diff
@@ -0,0 +1,844 @@
...
+    );
+    assertEq(
+      flexClient.getPastRawBalance(_user, block.number - 1),
+      _amountA + _amountB,
+      "user's second deposit was not properly checkpointed"
+    );
+  }
+}
+
+contract Vote is FlexVotingClientTest {
```
apbendi, line 215: It would be nice to break these tests up a bit and follow the naming conventions required for `scopelint spec` where possible. [Link](https://github.com/ScopeLift/scopelint?tab=readme-ov-file#scopelint-spec)

In particular, creating contracts for `ExpressVote` and `CastVote` and moving these `Vote` tests into those as appropriate seems like it'd be easy. We could also do a test for `Constructor` pretty easily from the current `Deploy` test contract.

In the longer run, for the next version of this lib, I think we should think a bit more about how we structure the tests and mocking. But this suite is already a _huge_ improvement, given we had no testing internal to this lib before!

apbendi (approved):
Looking good @davidlaprade. Left some comments that could be addressed, but nothing blocking.

.github/workflows/ci.yml
davidlaprade: Done in 12b33a6

foundry.toml
```diff
@@ -1,10 +1,8 @@
 [profile.default]
-  # We don't specify a solc version because Aave contracts are pinned to 0.8.10,
-  # but we use more recent solc versions for other contracts, so we let forge
-  # auto-detect solc versions.
   optimizer = true
   optimizer_runs = 10_000_000
   remappings = ["@openzeppelin/contracts/=lib/openzeppelin-contracts/contracts"]
+  solc_version = '0.8.25'
```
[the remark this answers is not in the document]
davidlaprade: Done in 2adaab1

test/FlexVotingClient.t.sol
```diff
@@ -0,0 +1,844 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.10;
```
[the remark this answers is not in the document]
davidlaprade: Yup, done in 1f1ffa8

test/FlexVotingClient.t.sol
```diff
@@ -0,0 +1,844 @@
...
+import {Vm} from "forge-std/Vm.sol";
+import {IVotes} from "@openzeppelin/contracts/governance/utils/IVotes.sol";
+import {IGovernor} from "@openzeppelin/contracts/governance/Governor.sol";
+
+import {IVotingToken} from "../src/interfaces/IVotingToken.sol";
+import {IFractionalGovernor} from "../src/interfaces/IFractionalGovernor.sol";
+import {MockFlexVotingClient} from "./MockFlexVotingClient.sol";
+import {GovToken} from "./GovToken.sol";
+import {FractionalGovernor} from "./FractionalGovernor.sol";
+import {ProposalReceiverMock} from "./ProposalReceiverMock.sol";
```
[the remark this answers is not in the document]
davidlaprade: Great idea, this is much nicer. Done in 45b3775

src/FlexVotingClient.sol
```diff
@@ -194,19 +199,21 @@ abstract contract FlexVotingClient {
 
   /// @dev Checkpoints the _user's current raw balance.
   function _checkpointRawBalanceOf(address _user) internal {
-    balanceCheckpoints[_user].push(_rawBalanceOf(_user));
+    balanceCheckpoints[_user].push(SafeCast.toUint48(block.number), _rawBalanceOf(_user));
```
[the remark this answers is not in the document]
davidlaprade: Will handle this in a follow up PR, thanks!

test/FlexVotingClient.t.sol
davidlaprade: Will handle this in a follow up PR, thanks!

src/FlexVotingClient.sol
davidlaprade: Switched to Trace208, thanks!

src/FlexVotingClient.sol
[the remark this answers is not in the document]
davidlaprade: Tracking [here](https://github.com/ScopeLift/flexible-voting/issues/70)

PR #71, opened 2024-11-13
Improve FlexVotingClient test structure

### LINKED ISSUE #69: Improve FlexVotingClient test structure
https://github.com/ScopeLift/flexible-voting/pull/66#discussion_r1838279620

image

davidlaprade:
Fixes #69

This PR restructures the FlexClient tests to accord with `scopelint spec`. Some tests were renamed. Other tests were added. The resulting spec is:

```
Contract Specification: FlexVotingClient
├── constructor
│   ├──  Sets Governor
│   └──  Self Delegates
├── _rawBalanceOf
│   ├──  Returns Zero For Non Depositors
│   ├──  Increases On Deposit
│   ├──  Decreases On Withdrawal
│   └──  Unaffected By Borrow
├── _castVoteReasonString
│   └──  Returns Descriptive String
├── _selfDelegate
│   └──  Sets Client As The Delegate
├── expressVote
│   ├──  Increments Internal Accouting
│   ├──  Revert When: Depositing After Proposal
│   ├──  Revert When: No Client Weight But Token Weight
│   ├──  Revert On: Double Votes
│   └──  Revert On: Unknown Vote Type
├── castVote
│   ├──  Submits Votes To Governor
│   ├──  Weight Is Snapshot Dependent
│   ├──  Tracks Multiple Users Votes
│   ├──  Scales Vote Weight Based On Pool Balance
│   ├──  Abandons Unexpressed Voting Weight
│   ├──  Voting Weight Is Unaffected By Deposits After Proposal
│   ├──  Can Call Multiple Times For The Same Proposal
│   ├──  Revert When: No Votes To Cast
│   └──  Revert When: After Voting Period
├── _checkpointRawBalanceOf
│   └──  Stores The Raw Balance With The Block Number
├── getPastRawBalance
│   ├──  Returns Zero For Users Without Deposits
│   ├──  Returns Current Value For Future Blocks
│   └──  Returns User Balance At A Given Block
└── getPastTotalBalance
    ├──  Returns Zero Without Deposits
    ├──  Returns Current Value For Future Blocks
    ├──  Sums All User Deposits
    └──  Returns Total Deposits At A Given Block
```

### DIFF
--- a/test/FlexVotingClient.t.sol
+++ b/test/FlexVotingClient.t.sol
@@ -20,6 +20,10 @@ contract FlexVotingClientTest is Test {
   FractionalGovernor governor;
   ProposalReceiverMock receiver;
 
+  // This max is a limitation of GovernorCountingFractional's vote storage size.
+  // See GovernorCountingFractional.ProposalVote struct.
+  uint256 MAX_VOTES = type(uint128).max;
+
   function setUp() public {
     token = new GovToken();
     vm.label(address(token), "token");
@@ -90,7 +94,7 @@ contract FlexVotingClientTest is Test {
     returns (uint208 _boundedWeight)
   {
     _assumeSafeUser(_account);
-    _boundedWeight = uint208(bound(_voteWeight, 1, type(uint128).max));
+    _boundedWeight = uint208(bound(_voteWeight, 1, MAX_VOTES));
   }
 
   function _assumeSafeVoteParams(address _account, uint208 _voteWeight, uint8 _supportType)
@@ -103,7 +107,7 @@ contract FlexVotingClientTest is Test {
     _boundedSupport = _randVoteType(_supportType);
 
     // This max is a limitation of the fractional governance protocol storage.
-    _boundedWeight = uint208(bound(_voteWeight, 1, type(uint128).max));
+    _boundedWeight = uint208(bound(_voteWeight, 1, MAX_VOTES));
   }
 }
 
@@ -139,7 +143,7 @@ contract _RawBalanceOf is FlexVotingClientTest {
 
   function testFuzz_IncreasesOnDeposit(address _user, uint208 _amount) public {
     _assumeSafeUser(_user);
-    _amount = uint208(bound(_amount, 1, type(uint128).max));
+    _amount = uint208(bound(_amount, 1, MAX_VOTES));
 
     // Deposit some gov.
     _mintGovAndDepositIntoFlexClient(_user, _amount);
@@ -149,7 +153,7 @@ contract _RawBalanceOf is FlexVotingClientTest {
 
   function testFuzz_DecreasesOnWithdrawal(address _user, uint208 _amount) public {
     _assumeSafeUser(_user);
-    _amount = uint208(bound(_amount, 1, type(uint128).max));
+    _amount = uint208(bound(_amount, 1, MAX_VOTES));
 
     // Deposit some gov.
     _mintGovAndDepositIntoFlexClient(_user, _amount);
@@ -163,7 +167,7 @@ contract _RawBalanceOf is FlexVotingClientTest {
 
   function testFuzz_UnaffectedByBorrow(address _user, uint208 _deposit, uint208 _borrow) public {
     _assumeSafeUser(_user);
-    _deposit = uint208(bound(_deposit, 1, type(uint128).max));
+    _deposit = uint208(bound(_deposit, 1, MAX_VOTES));
     _borrow = uint208(bound(_borrow, 1, _deposit));
 
     // Deposit some gov.
@@ -214,7 +218,7 @@ contract _CheckpointRawBalanceOf is FlexVotingClientTest {
   ) public {
     vm.assume(_user != address(flexClient));
     _blockNum = uint48(bound(_blockNum, block.number + 1, type(uint48).max));
-    _amount = uint208(bound(_amount, 1, type(uint128).max));
+    _amount = uint208(bound(_amount, 1, MAX_VOTES));
 
     flexClient.exposed_setDeposits(_user, _amount);
     assertEq(flexClient.getPastRawBalance(_user, _blockNum), 0);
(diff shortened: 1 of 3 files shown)


test/FlexVotingClient.t.sol
```diff
@@ -196,8 +440,12 @@ contract Deposit is FlexVotingClientTest {
   }
 }
 
-contract Vote is FlexVotingClientTest {
-  function testFuzz_UserCanCastVotes(address _user, uint208 _voteWeight, uint8 _supportType) public {
+contract ExpressVote is FlexVotingClientTest {
```
davidlaprade, line 443: Probably worth adding a test that tries to express votes on a bogus proposalId

test/FlexVotingClient.t.sol
```diff
@@ -102,6 +102,250 @@ contract Deployment is FlexVotingClientTest {
...
+}
+
+contract GetPastRawBalance is FlexVotingClientTest {
+  function testFuzz_ReturnsZeroForUsersWithoutDeposits(
+    address _depositor,
+    address _nonDepositor,
+    uint208 _amount
+  ) public {
+    vm.assume(_depositor != address(flexClient));
+    vm.assume(_nonDepositor != address(flexClient));
```
apbendi, line 223: I'm surprised we're not getting test failures here, because I think we would need an assumption that `_depositor != _nonDepositor`. That makes me wonder if something is not right with the test, and it's passing for the wrong reason?

test/FlexVotingClient.t.sol
```diff
@@ -102,6 +102,250 @@ contract Deployment is FlexVotingClientTest {
...
+
+// Contract name has a leading underscore for scopelint spec support.
+contract _CheckpointRawBalanceOf is FlexVotingClientTest {
+  function testFuzz_StoresTheRawBalanceWithTheBlockNumber(
+    address _user,
+    uint208 _amount,
+    uint48 _blockNum
+  ) public {
+    vm.assume(_user != address(flexClient));
+    vm.assume(_blockNum > 2);
```
apbendi, line 203: For numeric values, we should use [bound](https://book.getfoundry.sh/reference/forge-std/bound?highlight=bound#bound) rather than assume, for performance reasons. Also, it might be better to use `block.number` here instead of assuming the current block number is less than 2. We could end up changing `setUp` to roll the block number, for some reason, or the tests could end up running in a different context in the future (such as a fork test from a network).

test/FlexVotingClient.t.sol
```diff
@@ -102,6 +102,250 @@ contract Deployment is FlexVotingClientTest {
...
+    vm.assume(_blockNum < type(uint48).max);
+    _amount = uint208(bound(_amount, 1, type(uint128).max));
+
+    flexClient.exposed_setDeposits(_user, _amount);
+    assertEq(flexClient.getPastRawBalance(_user, _blockNum), 0);
+
+    vm.roll(_blockNum);
```
apbendi, line 210: It would be best to do `vm.roll(block.number + _blockNum)` here, or alternatively, we can bound the `_blockNum` value to something above `block.number` in the first place, as suggested above.

test/FlexVotingClient.t.sol
```diff
@@ -354,51 +565,68 @@ contract Vote is FlexVotingClientTest {
...
-    vm.assume(_supportType > uint8(GCF.VoteType.Abstain));
-
-    vm.assume(_user != address(flexClient));
-    // This max is a limitation of the fractional governance protocol storage.
-    _voteWeight = uint208(bound(_voteWeight, 1, type(uint128).max));
+contract CastVote is FlexVotingClientTest {
+  function testFuzz_SubmitsVotesToGovernor(address _user, uint208 _voteWeight, uint8 _supportType)
+    public
+  {
+    _voteWeight = _commonFuzzerAssumptions(_user, _voteWeight, _supportType);
```
apbendi, line 595: We should refactor this test helper for a couple reasons, but I will pull that into its own issue since it's not part of what this PR is changing.

test/FlexVotingClient.t.sol
```diff
@@ -102,6 +102,250 @@ contract Deployment is FlexVotingClientTest {
...
+  }
+
+  function test_SelfDelegates() public view {
+    assertEq(token.delegates(address(flexClient)), address(flexClient));
+  }
+}
+
+// Contract name has a leading underscore for scopelint spec support.
+contract _RawBalanceOf is FlexVotingClientTest {
+  function _commonUserAssumptions(address _user) internal view {
```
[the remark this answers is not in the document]
davidlaprade: I like that better, thanks. Done in 4fb9abc

test/FlexVotingClient.t.sol
davidlaprade: Great catch. You were right. Fixed in 34e4c2c

test/FlexVotingClient.t.sol
davidlaprade: Doh 🤦. Done 4d6d8b2

test/FlexVotingClient.t.sol
davidlaprade: Taken care of in 4d6d8b2

test/FlexVotingClient.t.sol
davidlaprade: Decided to just do it in here, see [11fe8d6](https://github.com/ScopeLift/flexible-voting/pull/71/commits/11fe8d67a2e7997ee6cf3b8f378cda67604a3d2d)

test/FlexVotingClient.t.sol
davidlaprade: Done in 6e2e970

test/FlexVotingClient.t.sol
apbendi: Nice I like the approach you took

apbendi (approved):
Looks good @davidlaprade thanks!

PR #72, opened 2024-11-15
Add FlexVotingClient Invariant Tests

davidlaprade:
I tried to keep the handler as simple as I could, but it's more complicated than I'd like.

To help me reason about the reverts/failures I was seeing in my invariant tests, I ended up needing to write tests for the handler functions. This helped me catch a lot of bugs in the handler, but it made this PR quite a bit bigger than I'd have liked.

If there are important invariants you think I've missed, please lmk! I'm sure there are some I haven't thought of.


test/handlers/FlexVotingClientHandler.sol
```diff
@@ -0,0 +1,235 @@
...
+
+  // TODO restrict expressVote to addresses that deposited BEFORE proposal.
+  function expressVote(
+    uint256 _proposalId,
+    uint8 _support,
+    uint256 _userSeed
+  ) useVoter(_userSeed) countCall("expressVote") external {
+    vm.assume(_proposals.length() > 0);
+
+    // TODO should we allow people to try to vote with bogus support types?
```
apbendi, line 196: Fine line but I would say no. That the `expressVote` method reverts on an invalid vote type is a property that can/should be tested sufficiently in unit tests. Invariant tests should primarily be about ensuring the integrity of the system as it is used.

test/handlers/FlexVotingClientHandler.sol
```diff
@@ -0,0 +1,235 @@
...
+    flexClient.expressVote(_proposalId, _support);
+
+    ghost_actorProposalVotes[currentActor][_proposalId] += 1;
+    if (ghost_actorProposalVotes[currentActor][_proposalId] > 1) {
+      ghost_doubleVoteActors += 1;
+    }
+  }
+
+  function castVote(uint256 _proposalId) countCall("castVote") external {
+    console2.log("david trying to cast vote");
```
apbendi, line 210: Can be removed?

test/FlexVotingClient.invariants.t.sol
```diff
@@ -0,0 +1,183 @@
...
+    assert(flexClient.deposits(address(this)) > 0);
+    assert(flexClient.deposits(address(this)) < _amount);
+  }
+
+  function testFuzz_expressVote(
+    uint256 _userSeed,
+    uint256 _proposalId,
+    uint128 _amount
+  ) public {
+    _amount = uint128(bound(_amount, 1, type(uint128).max));
```
apbendi, line 129: In invariant tests, I prefer to use `_bound` instead of `bound`. The former does *not* log the result of the bound to console, and_for some reason—forge seems to include the invariant test bound results in stdout for invariant tests even if they don't fail.

test/handlers/FlexVotingClientHandler.sol
```diff
@@ -0,0 +1,235 @@
...
+  // TODO we restrict withdrawals to addresses that have balances, should we?
+  function withdraw(
+    uint256 _userSeed,
+    uint208 _amount
+  )
+    useActor(_userSeed)
+    countCall("withdraw")
+    external
+  {
+    // TODO we limit withdrawals to the total amount deposited, should we?
```
apbendi, line 150: Sounding like a broken record here, but if we know a call will definitely revert every time, and we have that case sufficiently unit tested, I think it's ok to bound it out. Fine line, though, and this is a weakly held opinion. I have accidentally bounded out valid failure cases before.

apbendi (commented):
Hey @davidlaprade, I know this is in draft but I had some time to take a look and leave some preliminary thoughts. Looking good overall! Will be nice to have an invariant suite for this.

test/FlexVotingClient.invariants.t.sol
```diff
@@ -0,0 +1,183 @@
+// SPDX-License-Identifier: MIT
```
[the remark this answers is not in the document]
davidlaprade: Yeah, I wrote some handler tests then realized I needed to make substantial changes in order to get reverts to a reasonable level, which then broke my handler tests. They are now updated 👍

test/FlexVotingClient.invariants.t.sol
davidlaprade: Didn't know about that. Done! Thanks

test/handlers/FlexVotingClientHandler.sol
```diff
@@ -0,0 +1,235 @@
...
+
+  function _remainingTokens() internal returns (uint128) {
+    return type(uint128).max - ghost_mintedTokens;
+  }
+
+  function proposalLength() public returns(uint256) {
+    return _proposals.length();
+  }
+
+  // TODO This always creates a new actor. Should it?
```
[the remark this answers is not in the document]
davidlaprade: Good call, e7738b5

test/handlers/FlexVotingClientHandler.sol
```diff
@@ -0,0 +1,235 @@
...
+    ghost_withdrawSum += _amount;
+    ghost_accountDeposits[currentActor] -= uint128(_amount);
+  }
+
+  function propose(
+    string memory _proposalName,
+    uint256 _seed
+  ) countCall("propose") external {
+    // Proposal will underflow if we're on the zero block
+    if (block.number == 0) vm.roll(1);
```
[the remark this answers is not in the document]
davidlaprade: Done in 04638eb, thanks

davidlaprade:
Weirdly test coverage is ... down(?) because of a PR that only adds... tests 🤔

test/handlers/FlexVotingClientHandler.sol
```diff
@@ -0,0 +1,365 @@
...
+  uint8 public PROPOSAL_THRESHOLD = 70;
+
+  EnumerableSet.UintSet internal proposals;
+  EnumerableSet.AddressSet internal voters;
+  EnumerableSet.AddressSet internal actors;
+  address internal currentActor;
+
+  // Maps proposalIds to sets of users that have expressed votes but not yet had
+  // them cast.
+  mapping(uint256 => EnumerableSet.AddressSet) internal pendingVotes;
```
davidlaprade, line 39:
```suggestion
  mapping(uint256 => EnumerableSet.AddressSet) internal ghost_pendingVotes;
```

test/handlers/FlexVotingClientHandler.sol
```diff
@@ -0,0 +1,365 @@
...
+    _amount = uint208(_bound(_amount, 0, remainingTokens()));
+
+    // Some actors won't have the tokens they need. This is deliberate.
+    if (_amount <= remainingTokens()) {
+      token.exposed_mint(currentActor, _amount);
+      ghost_mintedTokens += uint128(_amount);
+    }
+
+    vm.startPrank(currentActor);
+    // TODO We're pre-approving every deposit, should we?
```
davidlaprade, line 199: I tried to call out the assumptions as I was making them. Feel free to push back if you think any of them were incorrect or too strong.

test/FlexVotingClient.invariants.t.sol
```diff
@@ -0,0 +1,140 @@
...
+
+    // Proposal will underflow if we're on the zero block.
+    if (block.number == 0) vm.roll(1);
+
+    bytes4[] memory selectors = new bytes4[](5);
+    selectors[0] = FlexVotingClientHandler.deposit.selector;
+    selectors[1] = FlexVotingClientHandler.propose.selector;
+    selectors[2] = FlexVotingClientHandler.expressVote.selector;
+    selectors[3] = FlexVotingClientHandler.castVote.selector;
+    selectors[4] = FlexVotingClientHandler.withdraw.selector;
```
davidlaprade, line 48: Do we want to expose a handler function to move the clock forward?

davidlaprade: Done in 3ed2b46

test/handlers/FlexVotingClientHandler.sol
davidlaprade: I don't think this is really a ghost variable, since it's internal-only. Ghost variables track state so that it can be compared at test time. That's not what this does.

alexkeating (approved):
LGTM thanks for adding the passage of time

PR #75, opened 2024-11-21
Switch back to nightly build of foundry

davidlaprade:
Remappings were causing the issue that broke CI a few days ago. Figured it out by making a minimal repro [here](https://github.com/davidlaprade/foundry-ci-failure-repro/actions/workflows/ci.yml)

This was caused by a bug in foundry, reported here https://github.com/foundry-rs/foundry/issues/9375 and fixed here https://github.com/foundry-rs/foundry/pull/9379


apbendi (approved):
LGTM!

PR #76, opened 2024-12-05
Switch to custom errors

### LINKED ISSUE #73: Convert require statements with errors to throw custom error types
Current best practice is to throw custom error types rather than strings on revert cases, and to append the contract name to the front of custom error types with `__` in between. We should convert all revert cases in the FlexibleVotingClient.sol to use this convention. Note that, except in the latest version of solidity,  you cannot throw custom error types in `require` statements. As a library, we need to support older Solidity versions, so we should also convert from `require` statements to conditionals.

As an example, the case:

```solidity
require(weight > 0, "no weight");
```

Might be converted to something like:

```solidity
if (weight == 0) revert FlexVotingClient__NoVotingWeight();
```

davidlaprade:
Fixes #73

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -106,7 +106,7 @@ jobs:
         uses: zgosalvez/github-actions-report-lcov@v2
         with:
           coverage-files: ./lcov.info
-          minimum-coverage: 92 # Set coverage threshold.
+          minimum-coverage: 91 # Set coverage threshold.
 
   slither-analyze:
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


alexkeating (approved):
Lgtm, fix the build before merging

davidlaprade:
Thanks!

Seems like something changed in [zgosalvez/github-actions-report-lcov](https://github.com/zgosalvez/github-actions-report-lcov) that reduced our coverage from 92.5% to 91.6%. I just ran master and it's broken [there too](https://github.com/ScopeLift/flexible-voting/actions/runs/12058053014). So I'm just going to reduce the target coverage number to get things green.

PR #82, opened 2025-01-15
Client sub delegation

davidlaprade:
Adds the ability to sub-delegate within a FlexVotingClient, i.e. to delegate one's internal voting weight to another user (who may or may not have voting weight within the client as well).

Sub delegation is an optional add-on via a new abstract contract `FlexVotingDelegatable`. To make this possible, shared storage vars and logic were abstracted into a new abstract contract `FlexVotingBase` which both `FlexVotingClient` and `FlexVotingDelegatable` inherit from. Some internal functions and variables were renamed accordingly.

Someone wishing to implement a client with sub-delegation will want a contract with (at minimum) the following inheritance structure:

```solidity
contract MyClient is FlexVotingBase, FlexVotingClient, FlexVotingDelegatable {
  // ...
}
```


davidlaprade:
I'm not sure why `scopelint check` is [failing on CI](https://github.com/ScopeLift/flexible-voting/actions/runs/12893921772/job/35951400997?pr=82). It's passing locally. My working directory is clean and I'm on the latest version of scopelint.

test/SharedFlexVoting.t.sol
```diff
@@ -0,0 +1,1292 @@
...
+    flexClient.exposed_checkpointTotalVoteWeight(_initBalance);
+
+    _advanceTimeTo(_timepoint);
+    flexClient.exposed_checkpointTotalVoteWeight(_delta);
+    _advanceTimeBy(1);
+
+    assertEq(flexClient.getPastTotalVoteWeight(_timepoint), uint256(_initBalance + _delta));
+  }
+
+  function testFuzz_RevertIf_negativeDeltaWraps(int256 delta, uint208 balance) public {
```
apbendi, line 298: Two things:

1. This test is failing for me with the error:

```
[FAIL: call didn't revert at a lower depth than cheatcode call depth; counterexample: calldata=0x461c6f850000000000000000000000000000000000000000000000000000000000003200000000000000000000000000000000000000000000000000000000002ade3881 args=[12800 [1.28e4], 719206529 [7.192e8]]] testFuzz_RevertIf_negativeDeltaWraps(int256,uint208) (runs: 0, μ: 0, ~: 0)
```

2. What is this test actually...testing? I don't actually see where any of our code is under test. What am I missing?

src/FlexVotingBase.sol
```diff
@@ -0,0 +1,126 @@
...
+    GOVERNOR = IFractionalGovernor(_governor);
+  }
+
+  /// @dev Returns a representation of the current amount of `GOVERNOR`s
+  /// token that `_user` has claim to in this system. It may or may not be
+  /// equivalent to the withdrawable balance of `GOVERNOR`s token for `user`,
+  /// e.g. if the internal representation of balance has been scaled down.
+  function _rawBalanceOf(address _user) internal view virtual returns (uint208);
+
+  // TODO rename to avoid collision with FlexVotingDelegatable.
```
apbendi: Flagging this: is it something we need to address in this PR? Or pull into an issue?

src/FlexVotingBase.sol
```diff
@@ -0,0 +1,126 @@
...
+  /// @dev Delegates the present contract's voting rights with `GOVERNOR` to itself.
+  function _selfDelegate() internal {
+    IVotingToken(GOVERNOR.token()).delegate(address(this));
+  }
+
+  function _applyDeltaToCheckpoint(Checkpoints.Trace208 storage _checkpoint, int256 _delta)
```
apbendi, line 90: Would be nice to have natspec for this method. Alternatively we could pull out an issue to do a natspec pass/review for the whole repo.

src/FlexVotingClient.sol
```diff
@@ -5,57 +5,22 @@ import {SafeCast} from "@openzeppelin/contracts/utils/math/SafeCast.sol";
...
-abstract contract FlexVotingClient {
-  using SafeCast for uint256;
-
-  // @dev Trace208 is used instead of Trace224 because the former allocates 48
-  // bits to its _key. We need at least 48 bits because the _key is going to be
-  // a timepoint. Timepoints in the context of ERC20Votes and ERC721Votes
-  // conform to the EIP-6372 standard, which specifies they be uint48s.
+/// This contract extends FlexVotingBase, adding two features:
+///   (a) the ability for depositors to express voting preferences on
+///       {Governor}'s proprosals, and
```
apbendi: nit: spelling of "proposals" is wrong

src/FlexVotingClient.sol
```diff
@@ -70,64 +35,38 @@ abstract contract FlexVotingClient {
...
-  // https://github.com/OpenZeppelin/openzeppelin-contracts/blob/7b74442c5e87ea51dde41c7f18a209fa5154f1a4/contracts/governance/extensions/GovernorCountingFractional.sol#L37
+  /// Constant used by OZ's implementation of {GovernorCountingFractional} to
+  /// signal fractional voting.
+  /// https://github.com/OpenZeppelin/openzeppelin-contracts/blob/7b74442c5e87ea51dde41c7f18a209fa5154f1a4/contracts/governance/extensions/GovernorCountingFractional.sol#L37
   uint8 internal constant VOTE_TYPE_FRACTIONAL = 255;
 
   error FlexVotingClient__NoVotingWeight();
   error FlexVotingClient__AlreadyVoted();
   error FlexVotingClient__InvalidSupportValue();
   error FlexVotingClient__NoVotesExpressed();
```
apbendi, line 51: nit: let's create a separate issue to review the repo and re-arrange declarations to our conventional ordering.

test/MockFlexVotingClient.sol
```diff
@@ -75,8 +81,9 @@ contract MockFlexVotingClient is FlexVotingClient {
...
     deposits[msg.sender] -= _amount;
 
-    FlexVotingClient._checkpointRawBalanceOf(msg.sender);
-    FlexVotingClient._checkpointTotalBalance(-1 * int256(uint256(_amount)));
+    int256 _delta = -1 * int256(uint256(_amount));
+    _checkpointVoteWeightOf(msg.sender, _delta);
+    _checkpointTotalVoteWeight(_delta);
 
     TOKEN.transfer(msg.sender, _amount); // Assumes revert on failure.
   }
```
apbendi, line 89: *arbitrary comment location*

I feel like I must be missing something. Reading the implementation of this Mock, I don't should the `borrow` method checkpoint the total voting weight?

src/FlexVotingDelegatable.sol
```diff
@@ -0,0 +1,84 @@
...
+///   - user A deposits 100 governance tokens in a FlexVotingClient
+///   - user B deposits 50 governance tokens into the same client
+///   - user A delegates voting weight to user B
+///   - a proposal is created in the Governor contract
+///   - user B expresses a voting preference P on the proposal to the client
+///   - the client casts its votes on the proposal to the Governor contract
+///   - user B's voting weight is combined with user A's voting weight so that
+///     150 tokens are effectively cast with voting preference P on behalf of
+///     users A and B.
+abstract contract FlexVotingDelegatable is Context, FlexVotingBase {
```
apbendi: Naming bikeshed: some googling seems to indicate "Delegatable" is not a real word but "Delegable" is. I'm not saying that definitely means we should change it, but I think we should consider it.

src/FlexVotingDelegatable.sol
```diff
@@ -0,0 +1,84 @@
...
+
+import {FlexVotingBase} from "src/FlexVotingBase.sol";
+
+/// @notice This is an abstract contract designed to make it easy to build
+/// clients for governance systems that inherit from GovernorCountingFractional,
+/// a.k.a. Flexible Voting governors.
+///
+/// This contract extends FlexVotingBase, adding the ability to subdelegate one's
+/// internal voting weight. It is meant to be inherited from in conjunction with
+/// FlexVotingClient. Doing so makes the following usecase possible:
```
apbendi: I know we discussed this before, so can you remind me: If this contract is meant to be used with `FlexVotingClient`, then why not inherit from `FlexVotingClient` directly, instead of from the base? Why make the user figure out to inherit from both.

apbendi (commented):
Not a totally full review @davidlaprade, but left some preliminary feedback. Will pick this back up quickly after your responses to continue the review cycle. Overall I like how it's put together but I feel like maybe I'm missing something so I want to get your response first (see, in particular, my question on why the mock client doesn't checkpoint the total weight in `borrow`. I feel like I must be confused about the model so I want to straighten that out in my head before leaving more feedback).

src/FlexVotingDelegatable.sol
davidlaprade: Ah, I found where we talked about this:

image

It is more modular, but it's modularity that's not really interesting, since there really isn't a usecase for `FlexVotingBase + FlexVotingDelegatable` without also including `FlexVotingClient`.

I do still kind of agree that it is more self-documenting to have `MyContract is FlexVotingClient, FlexVotingDelegatable` than it is to just do `MyContract is FlexVotingDelegatable`.

What do you think is more important in this case: simplicity (one contract to inherit from) or self-documentation?

test/SharedFlexVoting.t.sol
davidlaprade: Discussed on our call.

This failing test is fixed [here](https://github.com/ScopeLift/flexible-voting/pull/84/commits/85d103755846cb348dac9d769228863bb8820886#diff-8c5bded8edfaac000e1001d7b497bac6fbfa60c57ed078671c86c0de3ca21645R329-R333) in #84

test/MockFlexVotingClient.sol
davidlaprade: Discussed on our call but documenting here for posterity...

It's a little confusing, but I think the existing behavior is actually correct. `borrow` doesn't checkpoint the total voting weight because that storage var (i.e. `totalVoteWeightCheckpoints`) is really storing the the total vote weight _that has been deposited into the system_, not the actual vote weight of the system. We don't need to checkpoint the _actual_ vote weight that the client has access to, that's already done for us by the governor's token.

The scenario to keep in mind is this:
* user A deposits 50 GOV into client
* user B deposits 100 GOV into client
* user C borrows 50 GOV from client
* 100 GOV remains as the client's balance
* proposal P is created
* user B expresses his vote on P -- how much weight should his preference get?
* the weight of user B's preference should NOT be:
  * user B's deposits / GOV.balanceOf(client @ snapshot(P))
  * this would be: 100 / 100 -- i.e. it would give B _all_ of the voting weight
* instead, the weight of user B's preference should be:
  * user B's deposits / the total amount of GOV deposited into client
  * i.e. 100/150, or 2/3rds of client's voting weight at the time of P

src/FlexVotingBase.sol
davidlaprade: Probably a good idea to review the whole repo. Created https://github.com/ScopeLift/flexible-voting/issues/86

src/FlexVotingClient.sol
davidlaprade: Added https://github.com/ScopeLift/flexible-voting/issues/87, thanks

src/FlexVotingDelegatable.sol
davidlaprade: I'm fine with the change. I think they're equally clear and all else being equal I'd rather we use real words 😄 

Done in 3e16491

src/FlexVotingClient.sol
davidlaprade: Thanks, see cb90ea6

test/SharedFlexVoting.t.sol
davidlaprade: Decided to just pull the fix in here via 1e1d9bb

src/FlexVotingBase.sol
davidlaprade: Created https://github.com/ScopeLift/flexible-voting/issues/88 and updated the comment to reference the GH issue in 345804b

davidlaprade:
I'm not sure why the format step keeps failing. Everything passes locally with no changes. I'm on the latest foundry and scopelint versions. Even tried it with `FOUNDRY_PROFILE=ci`.

image

PR #89, opened 2025-04-15
Troubleshoot forge fmt CI failures

davidlaprade:
Failing here https://github.com/ScopeLift/flexible-voting/pull/82

Trying to figure out why. Does CI fail on a ~clean branch off of master?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -130,3 +130,29 @@ jobs:
         uses: github/codeql-action/upload-sarif@v2
         with:
           sarif_file: ${{ steps.slither.outputs.sarif }}
+
+  connect:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - name: Install Foundry
+      uses: foundry-rs/foundry-toolchain@v1
+      with:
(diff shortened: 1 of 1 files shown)


davidlaprade:
The issue was the foundry nightly warning. Disabling it fixed the problem.

I used [this action](https://github.com/lhotari/action-upterm) to create [a CI job](https://github.com/ScopeLift/flexible-voting/pull/89/commits/ee0b8493cefe08a785c198323afd1a75d4ce55ed) that would allow me to SSH into the box to debug things in an environment with the failure. Once in the box, I wrote a rust script similar to [the one that was failing](https://github.com/ScopeLift/scopelint/blob/a1441ebc269c62bb1eaff6f90161e2c1ebdf01af/src/check/validators/formatting.rs#L23) and cargo built it, then ran it to see what step was actually failing. It was the `forge fmt --check` step. The only "error" in that step was the warning, so I tried silencing the warning with `FOUNDRY_DISABLE_NIGHTLY_WARNING=1` and `scopelint check` started passing. This was the problem.

I've cherrypicked 385a775 onto #82 and will merge it there. Closing this out.

